PR #2, opened 2021-02-10
Create the mssql role

spetrosi:
Write the role
Add test playbooks

PR #10, opened 2021-04-06
Add the advanced functionality

spetrosi:
This PR adds the following functionality:

- enable or disable the SQL agent
- install or remove the `mssql-server-fts` package that provides full-text search.
- install or remove the `mssql-server-ha` package and enable or disable the `hadrenabled` setting
- enable or disable settings that improve performance on hosts that support Forced Unit Access (FUA) capability 
- configure MSSQL to encrypt connections using TLS certificates


README.md
```diff
@@ -139,6 +153,99 @@ You can find an example of the SQL file at `tests/sql_script.sql`.
...
+improve performance on hosts that support Forced Unit Access (FUA) capability.
+
+Only set this variable to `true` if your hosts are configured for FUA
+capability.
+
+When set to `true`, the role applies the following settings:
+
+* Set the `traceflag 3979 on` setting to enable trace flag 3979 as a startup
+parameter
+* Set the `control.alternatewritethrough` setting to `0`
```
richm, line 196: do you mean `1` here?

README.md
```diff
@@ -139,6 +153,99 @@ You can find an example of the SQL file at `tests/sql_script.sql`.
...
+control node.
+
+When you use this variable, the role copies the TLS certificate and private key
+files to MSSQL and configure MSSQL to use these files to encrypt connections.
+
+Default: `null`
+
+Type: `dict`
+
+#### `certificate`
```
richm, line 226: For the crypto parameter names, please use the conventions here - https://linux-system-roles.github.io/documentation/tls_crypto_parameter_and_key_names.html

if `certificate` is the name of an x509 certificate file on the controller to copy to the managed host, use `cert_src`

also - `private_key_src`

tasks/copy_certs.yml
```diff
@@ -0,0 +1,42 @@
...
+        state: file
+      register: __mssql_certificate_tempfile
+
+    - name: Copy the {{ item }} file to the host
+      copy:
+        src: "{{ item }}"
+        dest: "{{ __mssql_certificate_tempfile.path }}"
+        owner: mssql
+        group: mssql
+        mode: 0600
```
richm, line 14: Why use a tempfile?
Should there be a separate `cert_src` which is copied to `/etc/pki/tls/certs/` and `private_key_src` which is copied to `/etc/pki/tls/private/`?  If so, what should be the name of the files in `/etc/pki/tls/certs/` and `/etc/pki/tls/private/`?  does mssql expect a certain file name?  should the file name be something like `cert_src | basename` and `private_key_src | basename`?  Do you need to give the user the ability to specify the name of the file on the managed host e.g. by using `cert` and `private_key` as in https://linux-system-roles.github.io/documentation/tls_crypto_parameter_and_key_names.html ?

tasks/main.yml
```diff
@@ -292,7 +278,101 @@
...
+  vars:
+    __mssql_conf_setting: "sqlagent enabled"
+    __mssql_conf_setting_value: "{{ mssql_enable_sql_agent }}"
+  when: mssql_enable_sql_agent is not none
+
+- name: Install the {{ __mssql_server_fts_packages }} package
+  package:
+    name: "{{ __mssql_server_fts_packages }}"
+    state: "{{ 'present' if mssql_install_fts else 'absent' }}"
+  when: mssql_install_fts
```
richm, line 297: You can't use a "bareword" here - use `when: mssql_install_fts | bool`

tasks/main.yml
```diff
@@ -292,7 +278,101 @@
...
         state: absent
+
+- name: Configure the sqlagent setting
+  include_tasks: mssql_conf_setting.yml
+  vars:
+    __mssql_conf_setting: "sqlagent enabled"
+    __mssql_conf_setting_value: "{{ mssql_enable_sql_agent }}"
+  when: mssql_enable_sql_agent is not none
+
+- name: Install the {{ __mssql_server_fts_packages }} package
```
richm, line 293: I would suggest to use `Ensure the MS SQL FTS packages` for the `name`

tasks/main.yml
```diff
@@ -292,7 +278,101 @@
...
+  package:
+    name: "{{ __mssql_server_fts_packages }}"
+    state: "{{ 'present' if mssql_install_fts else 'absent' }}"
+  when: mssql_install_fts
+  notify: Restart the mssql-server service
+
+- name: Install and enable HA
+  when: mssql_enable_ha is not none
+  block:
+    - name: Install the {{ __mssql_server_ha_packages }} package
```
richm, line 303: Ensure the MS SQL HA packages

tasks/main.yml
```diff
@@ -292,7 +278,101 @@
...
+      command: >-
+        /opt/mssql/bin/mssql-conf traceflag 3979
+        {{ "on" if mssql_tune_for_fua_storage else "off" }}
+      when: '"3979" not in __mssql_get_traceflag.stdout'
+
+    - name: Configure the alternatewritethrough setting
+      include_tasks: mssql_conf_setting.yml
+      vars:
+        __mssql_conf_setting: "control alternatewritethrough"
+        __mssql_conf_setting_value: "false"
```
richm, line 334: should the value be `"{{ mssql_tune_for_fua_storage }}"`?

tasks/mssql_conf_setting.yml
```diff
@@ -0,0 +1,23 @@
+- name: Check if the {{ __mssql_conf_setting }} setting is enabled
+  command: >-
```
richm, line 2: you can't use `command` since you are piping, you'll have to use `shell`

however, I recommend using `command`, then parsing the output using ansible/jinja2 filters

you should also use the `quote` filter on the `command` arguments `__mssql_conf_setting` and `__mssql_conf_setting_value`

tasks/main.yml
```diff
@@ -292,7 +278,101 @@
...
+  include_tasks: mssql_conf_setting.yml
+  vars:
+    __mssql_conf_setting: "sqlagent enabled"
+    __mssql_conf_setting_value: "{{ mssql_enable_sql_agent }}"
+  when: mssql_enable_sql_agent is not none
+
+- name: Install the {{ __mssql_server_fts_packages }} package
+  package:
+    name: "{{ __mssql_server_fts_packages }}"
+    state: "{{ 'present' if mssql_install_fts else 'absent' }}"
```
nhosoi, line 296: If `mssql_install_fts` is other than `true`, `package` is not executed? I.e., there is no chance to have `state: absent`?

tasks/mssql_conf_setting.yml
nhosoi, line 2: Does `command` module work with `|`? I'd think it should be `shell`.

tasks/copy_certs.yml
spetrosi: mssql does not care about the names, I also see examples when people copied certs to different directories. I think mssql itself only needs to be pointed to the location of the cert and key. I think RHEL itself does some symlinks when you put the certs into this location. I think that the `cert and private_key` options are not needed, using the original names is enough.

README.md
spetrosi: According to MS docs it should be zero.
https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-performance-best-practices?view=sql-server-ver15#sql-server-and-forced-unit-access-fua-io-subsystem-capability
The same in Louis scripts: https://github.com/redhat-cop/sqlserver-coi/blob/master/scripts/mssql-install-demo/enterprise-storage.sh

tasks/main.yml
spetrosi: Should always be false, replied above.

tasks/main.yml
spetrosi: Well, I say that `mssql_install_fts` must be of type bool. Hence this login. true = install, false = remove

tasks/mssql_conf_setting.yml
spetrosi: It worked for me, but Rich said the same, I will change this to shell.

tasks/main.yml
nhosoi: But, `when: false` (mssql_install_fts is set to false), this task won't be executed, will it? 
E.g., if you run this playbook,
```
  vars:
    mssql_install_fts: false

  tasks:
    - name: print
      debug:
        msg: "PRINTED - {{ mssql_install_fts }}"
      when: mssql_install_fts
```
this print task is skipped:
```
TASK [print] ***********************************************************************************************
skipping: [localhost]
```
So, there is no chance for the package to be removed. I'd think the condition has to be: `mssql_install_fts` is set (not none) and the boolean value is true or false?

tasks/main.yml
spetrosi: Ok, will do

tasks/main.yml
spetrosi: Will do.

tasks/mssql_conf_setting.yml
spetrosi: Ok will do

tasks/main.yml
spetrosi: Yes I see what is the issue now, I should change `when: mssql_install_fts` to `when: mssql_install_fts is not none`.

README.md
spetrosi: Done.

tasks/copy_certs.yml
spetrosi: Customization is something that we can always add later. I think that what the users transferring from windows require now is as little customization as possible, hence I am trying to make it simpler.

tasks/main.yml
spetrosi: Changed this to `is not none` as per Noriko's suggestion to make this work for both install and uninstall

tasks/main.yml
spetrosi: Done

tasks/main.yml
spetrosi: Done

tasks/mssql_conf_setting.yml
spetrosi: Changed this to not pipe and instead use ansible's regex_replace.
You are right about the quotes filter, I forgot to make this change with the last push. I've committed it locally and will push when the current tests pass.

tasks/main.yml
spetrosi: Done

tasks/mssql_conf_setting.yml
spetrosi: I got rid of piping as Rich suggested.

spetrosi:
@richm @nhosoi, I am facing a weird issue with the tests. If you look at the ansible log in the details of [rhel-8-y/ansible-2.8](https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-mssql-pull-linux-system-roles_mssql-10-4f143c9-rhel-8-y-20210416-080954/artifacts/test.log.html) test, at the last task `Verify the traceflag setting` and debugs prior to it.
Ansible failed to properly compare two conditions each of which are true. Ansible 2.9 tests all work fine. And I am not able to reproduce this locally neither with Ansible 2.9 nor 2.8. The same test on the same version as in the CI - Ansible 2.8.19 - works just fine:
```

TASK [Get the value of the traceflag setting] ************************************************************************************************************************************************
ok: [127.0.0.3]

TASK [Print Ansible version] *****************************************************************************************************************************************************************
ok: [127.0.0.3] => {
    "ansible_version.full": "2.8.19"
}

TASK [Print Ansible version] *****************************************************************************************************************************************************************
ok: [127.0.0.3] => {
    "ansible_version.string": "2.8.19"
}

TASK [Print vars to debug weird test failures in the CI] *************************************************************************************************************************************
ok: [127.0.0.3] => {
    "__mssql_conf_get_traceflag.stdout": "traceflag0 : 3979"
}

TASK [Print vars to debug weird test failures in the CI] *************************************************************************************************************************************
ok: [127.0.0.3] => {
    "'3979' in __mssql_conf_get_traceflag.stdout": true
}

TASK [Print vars to debug weird test failures in the CI] *************************************************************************************************************************************
ok: [127.0.0.3] => {
    "__verify_mssql_is_tuned_for_fua": true
}

TASK [Print vars to debug weird test failures in the CI] *************************************************************************************************************************************
ok: [127.0.0.3] => {
    "('3979' in __mssql_conf_get_traceflag.stdout) == __verify_mssql_is_tuned_for_fua": true
}

TASK [Print vars to debug weird test failures in the CI] *************************************************************************************************************************************
ok: [127.0.0.3] => {
    "(\"3979\" in __mssql_conf_get_traceflag.stdout) == __verify_mssql_is_tuned_for_fua": true
}

TASK [Verify the traceflag setting] **********************************************************************************************************************************************************
ok: [127.0.0.3] => {
    "changed": false,
    "msg": "All assertions passed"
}
```
Do you have any idea why this might happen?
I have added a last piece of functionality for this PR - to disable tls encryption, feel free to review.

spetrosi:
Pavel pointed me to the issue that has been causing the failure with comparison: https://github.com/pallets/jinja/issues/813
It has been fixed in newer versions of Jinja but not in the Jinja that the CI uses. I will add a workaround for this issue in this PR, and later replace the workaround with a proper solution in #13

spetrosi:
[citest]

spetrosi:
[citest]

tests/tasks/verify_settings.yml
```diff
@@ -152,29 +152,15 @@
...
-
-    - name: Print vars to debug weird test failures in the CI
-      debug:
-        var: >-
-          ('3979' in __mssql_conf_get_traceflag.stdout) ==
-          (__verify_mssql_is_tuned_for_fua)
+    - name: Verify the traceflag setting
+      assert:
+        that: "'3979' in __mssql_conf_get_traceflag.stdout"
+      when: __verify_mssql_is_tuned_for_fua
```
richm, line 158: `when: __verify_mssql_is_tuned_for_fua | bool` to avoid the bareword problem

or - if `__verify_mssql_is_tuned_for_fua` can be `null` - use 
```yaml
when:
  - __verify_mssql_is_tuned_for_fua is not none
  - __verify_mssql_is_tuned_for_fua | bool
```

spetrosi:
citest pending

spetrosi:
[citest pending]

tasks/main.yml
nhosoi: > Changed this to `is not none` as per Noriko's suggestion to make this work for both install and uninstall

Thank you, @spetrosi! How about adding `| bool` to the above line to ensure the value of mssql_install_fts is a valid boolean like this?
```
state: "{{ 'present' if mssql_install_fts | bool else 'absent' }}"
```

spetrosi: Users must provide a boolean value with this `mssql_install_fts` variable, it is written in README. And I want to force users to do this because tasks depend on this. Now if a user provides e.g. a string "no", instead of boolean, the task will fail with a descriptive error message forcing users to use booleans:
```
fatal: [127.0.0.3]: FAILED! => {"changed": true, "cmd": ["/opt/mssql/bin/mssql-conf", "set", "sqlagent", "enabled", "no"], "delta": "0:00:00.058723", "end": "2021-04-19 04:38:53.632981", "failed_when_result": true, "rc": 0, "start": "2021-04-19 04:38:53.574258", "stderr": "", "stderr_lines": [], "stdout": "Validation error on setting 'sqlagent.enabled'\nenabled value must be '1' or '0'", "stdout_lines": ["Validation error on setting 'sqlagent.enabled'", "enabled value must be '1' or '0'"]}
```
If I add | bool there, the task will just do `absent` if a user provides non-boolean.

tests/tasks/verify_settings.yml
spetrosi: Will apply `bool` there. This is an internal to role task, so I expect developers who will use this task to set to either `true` or `false`, and not to `none`.

tasks/main.yml
```diff
@@ -340,7 +340,7 @@
         __mssql_conf_setting_value: "{{ mssql_tune_for_fua_storage }}"
 
 - name: Configure MSSQL to encrypt connections
-  when: mssql_tls.enable
+  when: mssql_tls.enable | bool
```
richm, line 351: I think this is ok - because this is an operation (the `.` is a dereference operation) it will be evaluated correctly - but it is ok to leave this with the `| bool` for now

tests/tasks/verify_settings.yml
```diff
@@ -193,70 +208,79 @@
...
+    - name: Verify if the tlscert setting is set to mssql_tls.cert
       assert:
-        that: >-
-          (mssql_tls.cert | string in
-          __mssql_conf_get_tlscert.stdout)
-          == __verify_mssql_is_tsl_encrypted
-      when: mssql_tls.cert is defined
+        that: mssql_tls.cert | string in __mssql_conf_get_tlscert.stdout
+      when:
+        - __verify_mssql_is_tsl_encrypted | bool
```
richm, line 215: tls

tests/tasks/verify_settings.yml
```diff
@@ -193,70 +208,79 @@
...
+        - mssql_tls.cert is defined
 
     - name: Verify if the tlscert setting is unset
       assert:
         that: "'No setting for the given' in __mssql_conf_get_tlscert.stdout"
-      when: mssql_tls.cert is not defined
+      when:
+        - not __verify_mssql_is_tsl_encrypted | bool
```
richm, line 222: tls

tests/tasks/verify_settings.yml
```diff
@@ -193,70 +208,79 @@
...
-    - name: Verify the tlskey setting
+    - name: Verify if the tlskey setting is set to mssql_tls.private_key
       assert:
-        that: >-
-          (mssql_tls.private_key in __mssql_conf_get_tlskey.stdout)
-          == __verify_mssql_is_tsl_encrypted
-      when: mssql_tls.private_key is defined
+        that: mssql_tls.private_key in __mssql_conf_get_tlskey.stdout
+      when:
+        - __verify_mssql_is_tsl_encrypted | bool
```
richm, line 235: tls

tests/tasks/verify_settings.yml
```diff
@@ -193,70 +208,79 @@
...
+        - mssql_tls.private_key is defined
 
     - name: Verify that the tlskey setting is unset
       assert:
         that: "'No setting for the given' in __mssql_conf_get_tlskey.stdout"
-      when: mssql_tls.cert is not defined
+      when:
+        - not __verify_mssql_is_tsl_encrypted | bool
```
richm, line 242: tls

tests/tasks/verify_settings.yml
```diff
@@ -193,70 +208,79 @@
...
+    - name: Verify if the tlsprotocols setting is set to mssql_tls.tls_version
       assert:
         that: >-
-          (mssql_tls.tls_version | string in
-          __mssql_conf_get_tlsprotocols.stdout) ==
-          __verify_mssql_is_tsl_encrypted
-      when: mssql_tls.tls_version is defined
+          mssql_tls.tls_version | string in __mssql_conf_get_tlsprotocols.stdout
+      when:
+        - __verify_mssql_is_tsl_encrypted | bool
```
richm, line 256: tls

tests/tasks/verify_settings.yml
```diff
@@ -193,70 +208,79 @@
...
+        - mssql_tls.tls_version is defined
 
     - name: Verify if the tlsprotocols setting is unset
       assert:
         that: >-
           'No setting for the given' in __mssql_conf_get_tlsprotocols.stdout
-      when: mssql_tls.tls_version is not defined
+      when:
+        - not __verify_mssql_is_tsl_encrypted | bool
```
richm, line 264: tls

tests/tasks/verify_settings.yml
```diff
@@ -193,70 +208,79 @@
...
 
-    - name: Verify the forceencryption setting
+    - name: Verify if the forceencryption setting is true
+      vars:
+        __mssql_conf_get_forceencryption_value: >-
+          {{ __mssql_conf_get_forceencryption.stdout |
+          regex_replace( '^.*\s:\s', '') }}
+      assert:
+        that: __mssql_conf_get_forceencryption_value | bool
+      when: __verify_mssql_is_tsl_encrypted | bool
```
richm, line 280: tls

tests/tasks/verify_settings.yml
```diff
@@ -193,70 +208,79 @@
...
+    - name: Verify if the forceencryption setting is unset
       assert:
         that: >-
-          (__mssql_conf_get_forceencryption.stdout | bool ==
-          __verify_mssql_is_tsl_encrypted) or
-          ("No setting for the given" in
-          __mssql_conf_get_forceencryption.stdout ==
-          __verify_mssql_is_tsl_encrypted)
+          'No setting for the given' in __mssql_conf_get_forceencryption.stdout
+      when: not __verify_mssql_is_tsl_encrypted | bool
```
richm, line 286: tls

README.md
richm: So it is the same as the default value?

README.md
```diff
@@ -139,6 +153,112 @@ You can find an example of the SQL file at `tests/sql_script.sql`.
...
+the existing certificate and private key files if this variable is set to
+`false`.
+
+Default: `null`
+
+Type: `bool`
+
+#### `cert`
+
+Path to the certificate file to copy to MSSQL.
```
richm, line 241: This is the path and filename?

tasks/copy_certs.yml
richm: Ok - then you don't need the tempfile - you can use the `copy` module with the `src` is the `mssql_tls.cert`, and the dest is `/etc/pki/tls/certs/{{ mssql_tls.cert | basename }}`   https://docs.ansible.com/ansible/2.8/user_guide/playbooks_filters.html#id8 - you have to use `basename` because `mssql_tls.cert` could be a path and filename.

using `copy` makes it idempotent - it won't `copy` if the file already exists (unless you use `force` - but I'm not sure how to expose that to the user interface - a way for the user to specify force overwrite of existing files)

tasks/main.yml
```diff
@@ -292,7 +278,133 @@
...
+      with_items:
+        - "{{ mssql_tls.cert }}"
+        - "{{ mssql_tls.private_key }}"
+
+    - name: Configure the tlscert setting
+      include_tasks: mssql_conf_setting.yml
+      vars:
+        __mssql_conf_setting: "network tlscert"
+        __mssql_conf_setting_value: >-
+          /etc/pki/tls/certs/{{ mssql_tls.cert }}
```
richm, line 364: `           /etc/pki/tls/certs/{{ mssql_tls.cert | basename }}`

tasks/main.yml
```diff
@@ -292,7 +278,133 @@
...
+
+    - name: Configure the tlskey setting
+      include_tasks: mssql_conf_setting.yml
+      vars:
+        __mssql_conf_setting: "network tlskey"
+        __mssql_conf_setting_value: >-
+          /etc/pki/tls/private/{{ mssql_tls.private_key }}
```
richm, line 371: `           /etc/pki/tls/private/{{ mssql_tls.private_key | basename }}`

tasks/mssql_conf_setting.yml
```diff
@@ -0,0 +1,37 @@
...
+    /opt/mssql/bin/mssql-conf get {{ __mssql_conf_setting }}
+  ignore_errors: true
```
richm, line 4: Under what conditions will this command issue an error?  When the setting is not set?

tests/tests_tls.yml
```diff
@@ -0,0 +1,69 @@
...
+            - mssql.key
+
+    - name: Flush handlers
+      meta: flush_handlers
+
+    - name: Verify connectivity and settings
+      include_tasks: tasks/verify_settings.yml
+      vars:
+        __verify_mssql_password: "p@55w0rD"
+        __verify_mssql_is_tsl_encrypted: true
```
richm, line 52: tls

tests/tests_tls.yml
```diff
@@ -0,0 +1,69 @@
...
+          enable: false
+
+    - name: Flush handlers
+      meta: flush_handlers
+
+    - name: Verify connectivity and settings
+      include_tasks: tasks/verify_settings.yml
+      vars:
+        __verify_mssql_password: "p@55w0rD"
+        __verify_mssql_is_tsl_encrypted: false
```
richm, line 69: tls

tasks/main.yml
spetrosi: I am confused about this bareword issue. I expect the value of `mssql_tls.enable` to be a boolean, do I need to add `| bool` in that case? I'd rather wish the role to fail if the user provides NOT a boolean value. I type this in the readme. If I only expect boolean, a string `"true"` (not a bool `true`) won't work, but I think it is fine.

tasks/main.yml
spetrosi: Thanks a lot for this fix, I implemented it.

tasks/mssql_conf_setting.yml
spetrosi: I checked and `mssql-conf get` does not return errors. I tried getting a defined, undefined, and not existing setting and it never returned `1`. So I am removing ignore_errors from all `get` tasks, thank you for this catch.

tests/tests_tls.yml
spetrosi: lol fixed this everywhere

tasks/main.yml
richm: > I am confused about this bareword issue.

It happens when you use a boolean valued variable in a `when` or `assert` `that` by itself - no operations, no filters - that is, the variable is a bare word:
```
when: somevalue
```
I can't remember the reason that Ansible/yaml/jinja cannot parse this correctly as a boolean value.
If the word is not "bare" - if you use another conditional like `not`, or use a dereference operator, or a filter, then Ansible/yaml/jinja are able to parse it correctly.  So all of these are ok - they are not "bare":
```
when: mssql_tls.enable  # dereference operator

when: not somevalue  # uses "not"

when: somevalue | bool  # uses a filter
```

> I expect the value of `mssql_tls.enable` to be a boolean, do I need to add `| bool` in that case?

The only other case where you would use `| bool` is when the value of `mssql_tls.enable` is a string that you must convert to a boolean value.

> I'd rather wish the role to fail if the user provides NOT a boolean value. I type this in the readme. If I only expect boolean, a string `"true"` (not a bool `true`) won't work, but I think it is fine.

Ok - if `mssql_tls.enable` can be a string `"true"` or `"false"`, then you should use `| bool` here.

spetrosi: Got it, thank you for the explanation. I think it is fine to keep `| bool` in this case, and in all other cases when checking booleans for the sake of consistency. `| bool` doesn't break anything.

README.md
spetrosi: Ansible accepts both the filename, given that the file resides in the same directory as the ansible-playbook, and the full path. Do you suggest rewording this?

richm: > Do you suggest rewording this?

No - I just checked and this wording is consistent with other roles

README.md
richm: That is - the default value is `0` - setting `mssql_tune_for_fua_storage: true` will set it to `0` - so why set `control.alternatewritethrough` at all?  Can it be set to any value other than `0`?

tasks/copy_certs.yml
richm: I still don't understand why you are using a tempfile here.

Why not do this?

```yaml
    - name: Copy the {{ item }} file to the host
      copy:
        src: "{{ item }}"
        dest: /etc/pki/tls/{{ 'private' if item == mssql_tls.private_key else 'certs' }}/{{ __mssql_cert_tempfile.path | basename }}
        owner: mssql
        group: mssql
        mode: 0600
```

tasks/main.yml
richm: Is there some way it can be set to `true`?  If not, then can we just ignore this setting?

tasks/mssql_conf_setting.yml
```diff
@@ -0,0 +1,36 @@
...
+    (__mssql_conf_setting_value != __mssql_conf_get_value | bool)))
+  register: __mssql_conf_set
+  failed_when: >
+    ("error" in __mssql_conf_set.stdout | lower) or (__mssql_conf_set.rc == 1)
+  notify: Restart the mssql-server service
+
+- name: Unset the {{ __mssql_conf_setting }} setting
+  command: /opt/mssql/bin/mssql-conf unset {{ __mssql_conf_setting }}
+  when:
+    - '__mssql_conf_setting_value == "unset"'
```
richm, line 31: you don't need the single quotes here - you can just use
```yaml
     - __mssql_conf_setting_value == "unset"
```

tasks/mssql_conf_setting.yml
```diff
@@ -0,0 +1,36 @@
...
+    - '"No setting for the given" not in __mssql_conf_get_setting.stdout'
```
richm, line 32: You do need the single quotes here since the first argument is a quoted string value

tasks/mssql_conf_setting.yml
```diff
@@ -0,0 +1,36 @@
...
+  register: __mssql_conf_set
+  failed_when: >-
+    ("error" in __mssql_conf_set.stdout | lower) or (__mssql_conf_set.rc == 1)
```
richm, line 35:
```
     ("error" in __mssql_conf_set.stdout | lower) or (__mssql_conf_set is failed)
```
use the task result tests instead of comparison to `rc` directly - https://docs.ansible.com/ansible/2.8/user_guide/playbooks_tests.html#task-results

tasks/mssql_conf_setting.yml
spetrosi: I removed the single quotes here and it caused and error:
```
fatal: [127.0.0.3]: FAILED! => {"reason": "We were unable to read either as JSON nor YAML, these are the errors we got from each:\nJSON: Expecting value: line 1 column 1 (char 0)\n\nSyntax Error while loading YAML.\n  did not find expected '-' indicator\n\nThe error appears to be in '/home/spetrosi/Documents/work/linux-system-roles/mssql/tasks/mssql_conf_setting.yml': line 32, column 34, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n    - __mssql_conf_setting_value == \"unset\"\n    - \"No setting for the given\" not in __mssql_conf_get_setting.stdout\n                                 ^ here\n"}
```

README.md
spetrosi: Even though `0` is the default, users might have it set to `1`, for some internal reasons. In this case, the role must set it back to `0`.
I do not really know how to handle disabling this `mssql_tune_for_fua_storage`. For now, I just wrote what the role does explicitly to be fair.
Indeed the role always sets this setting to `0`. Another way would be to usnet this setting. Unsetting means setting the default value - again `0`. I checked that the default for this setting is `0` with Amit. Explicitly setting to `0` is more clear than unsetting then, and produces the same result. Do you have an idea how to improve it?

tasks/mssql_conf_setting.yml
richm: Right - you need the single quotes on the second condition because it starts with a double quoted string

tasks/main.yml
```diff
@@ -322,7 +322,7 @@
 
     - name: Set the 3979 traceflag
       command: >-
-        /opt/mssql/bin/mssql-conf traceflag 3979 "on"
+        /opt/mssql/bin/mssql-conf traceflag 3979 'on'
```
richm, line 325: I don't think you need quotes here at all

tasks/main.yml
```diff
@@ -346,68 +347,44 @@
...
 
     - name: Configure the tlscert setting
       include_tasks: mssql_conf_setting.yml
       vars:
+        __mssql_tls_cert_dest: >-
+          /etc/pki/tls/certs/{{ mssql_tls_cert | basename }}
         __mssql_conf_setting: "network tlscert"
         __mssql_conf_setting_value: >-
-          /etc/pki/tls/certs/{{ mssql_tls_cert | basename }}
+          {{ __mssql_tls_cert_dest if mssql_tls_enable else 'unset' }}
```
richm, line 368: nice!

tasks/copy_certs.yml
richm: I still don't understand what is the purpose of using `tempfile`.  Can you do this?
```yaml
- name: Copy the {{ item | basename }} TLS file to the appropriate location
  copy:
    src: "{{ item }}"
    dest: >-
      /etc/pki/tls/{{ 'private' if item == mssql_tls_private_key
      else 'certs' }}/{{ item | basename }}
    owner: mssql
    group: mssql
    mode: 0600
    force: "{{ mssql_tls_force }}"
```

spetrosi: Oh, you mean tempfile at all. You recommended me using this in the previous PR that introduced the roles, these lines:
https://github.com/linux-system-roles/mssql/pull/2/files#diff-3d0ff1709ca48add100327bb2a468e6c508fb92a159c64c4f99ad1df89d9bddeR263
I though that this is a best practice rule. I can copy directly too.

richm: For the `mssql_input_sql_file` - you need to copy it somewhere on the managed host, and there isn't a "proper" location for it like `/etc/pki/tls/certs/` for TLS certificate files.  So the best way to handle it is to copy it to the remote host in a tempfile so that you can use it with `sqlcmd`.

spetrosi: ooooh I see, thank you for the clarification, now it makes perfect sense. I will remove this tempfile block and just use the copy module tomorrow.

tasks/main.yml
spetrosi: true, fixed

tasks/copy_certs.yml
spetrosi: Fixed now

richm:
looks good - please squash and I think we're ready to merge

spetrosi:
> looks good - please squash and I think we're ready to merge

Squashing and pushing means that the PR will reserve the CI for ~5 hours. Can I merge using the *Squash and Merge* button instead of squashing manually and force-pushing?

richm (approved):
yes - please use squash and merge

PR #14, opened 2021-04-21
Removing unnecessary tests/setup_module_utils.sh.

nhosoi:
The local setup_module_utils.sh is no longer being used.


richm:
nack - just rm tests/setup_module_utils.sh from mssql

nhosoi:
> nack - just rm tests/setup_module_utils.sh from mssql

Ah, right. it's not needed any more... :p

richm (approved):
lgtm

PR #16, opened 2021-05-03
Print the time that task in the CI tests last

spetrosi:
The CI tests take 5h total because each test downloads the mssql packages again each time.

### DIFF
--- a/tests/ansible.cfg
+++ b/tests/ansible.cfg
@@ -1,0 +1,2 @@
+[defaults]
+callback_whitelist = profile_tasks


spetrosi:
Yay, it worked! @richm try looking at some of the CI ansible logs for `PLAY RECAP`, under each play Ansible prints time taken by the play task in descending order. The task that installs the mssql-server package sometimes takes 74.72s, sometimes 300s, sometimes 703.34s (e.g. in centos-7/ansible-2.8). Also Ansible sometimes prints not all tasks, sounds like from tests_idempotency it does not print time for all imported roles, which is weird.
That proves that packages download takes the most time, so setting up the proxy is the correct way forward.

spetrosi:
Also, the time measurements after each PLAY RECAP only display results for the latest inclusion of the role, so in the cases where tests input the role several times, e.g. in tests_idempotency, the time measurements do not show the long-taking tasks.

richm:
>  That proves that packages download takes the most time, so setting up the proxy is the correct way forward.

Not exactly.  It proves that the `package` module can take up to 700 seconds.  The package module is doing
- downloading rpm metadata
- updating rpm metadata
- downloading packages
- installing packages

We still don't know how much of that time exactly is downloading packages.

richm (approved):
lgtm

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

PR #19, opened 2021-06-09
Add mssql_install_powershell, tests, verification tests, update README

spetrosi:


tasks/main.yml
```diff
@@ -297,6 +297,13 @@
   when: mssql_install_fts is not none
   notify: Restart the mssql-server service
 
+- name: Ensure the {{ __mssql_powershell_packages }} package
+  package:
+    name: "{{ __mssql_powershell_packages }}"
+    state: "{{ 'present' if mssql_install_powershell | bool else 'absent' }}"
+  when: mssql_install_powershell is not none
+  notify: Restart the mssql-server service
```
richm, line 305: I don't think installing this package requires a server restart

tests/tasks/verify_settings.yml
```diff
@@ -104,21 +104,40 @@
           __verify_mssql_agent_is_enabled | string | lower in
           __mssql_conf_get_sqlagent.stdout | lower
 
-- name: Verify the mssql-server-fts package
+- name: Verify the {{ __mssql_server_fts_packages }} package
```
richm, line 107: It's generally not good to have unrelated changes in a PR - it would be better to have these sort of "clean up" commits in a separate PR

tasks/main.yml
spetrosi: You are right, fixed

tests/tasks/verify_settings.yml
spetrosi: Got you, I've moved the unrelated changes to  #21.

spetrosi:
@richm do you think that it is fine that the role removes packages if users provide a variable, e.g.  `mssql_install_powershell: no`? I think that the ability to both install and remove packages using the role is a good thing. I think that the problem is in the confusing variables names.
Maybe change `mssql_install_powershell` to e.g. `mssql_package_powershell`? Do you have better ideas?
It didn't confuse me in the first place but now watching with fresh eyes I think that this might be an issue.
Thank you

tests/tasks/verify_settings.yml
```diff
@@ -121,6 +121,24 @@
...
+    - name: Verify if the {{ __mssql_powershell_packages }} package is installed
+      assert:
+        that: __mssql_powershell_packages in ansible_facts.packages
+      when: __verify_mssql_powershell_is_installed | bool
+
+    - name: >
+        Verify if the {{ __mssql_powershell_packages }} package is not installed
+      assert:
+        that: __mssql_powershell_packages not in ansible_facts.packages
+      when: not __verify_mssql_fts_is_installed | bool
```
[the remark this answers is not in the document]
spetrosi: Good catch, fixed

nhosoi:
> @richm do you think that it is fine that the role removes packages if users provide a variable, e.g. `mssql_install_powershell: no`? I think that the ability to both install and remove packages using the role is a good thing. I think that the problem is in the confusing variables names.
> Maybe change `mssql_install_powershell` to e.g. `mssql_package_powershell`? Do you have better ideas?
> It didn't confuse me in the first place but now watching with fresh eyes I think that this might be an issue.
> Thank you

To my even fresher eyes, `mssql_install_powershell` looks good to me. If `mssql_install_powershell: false|no`, removing the package sounds natural to me. So, you have my +1 to `mssql_install_powershell`. :)

nhosoi (approved):
lgtm

richm:
```
fatal: [/cache/fedora-33.qcow2]: FAILED! => {"changed": false, "failures": [], "msg": "Unknown Error occured: Transaction test error:\n  installing package powershell-7.1.3-1.centos.8.x86_64 needs 26MB more space on the / filesystem\n\nError Summary\n-------------\nDisk Requirements:\n   At least 26MB more space needed on the / filesystem.\n", "rc": 1, "results": []}
```
26MB!  good grief

richm (approved):
lgtm
let's ignore the 'out of disk space' errors for now - I don't think there is an easy way to fix this with qemu/fmf - when we move to baseos CI we will have more disk provisioning options

PR #20, opened 2021-06-09
Remove the tests/.fmf directory, it is redundant and breaks rpmlint

spetrosi:


pcahyna:
are you sure it is redundant? I suspect that without it, provision.fmf will not work.

spetrosi:
This file is required, I will need to keep it within RPMbuild and ignore the check

PR #22, opened 2021-06-25
Support ver 2017

spetrosi:


tasks/main.yml
```diff
@@ -20,34 +20,76 @@
...
+  when: >-
+    (__mssql_server_packages not in ansible_facts.packages) or
+    (mssql_upgrade | bool)
 
-- name: Ensure that the Microsoft SQL Server package is installed
+- name: Ensure the {{ __mssql_server_packages }} package
   package:
-    name: "{{ __mssql_server_packages }}"
-    state: present
+    name: "{{ __mssql_server_packages }}-1*"
```
richm, line 86: ?

tasks/main.yml
```diff
@@ -20,34 +20,76 @@
...
+    state: "{{ 'latest' if mssql_upgrade else 'present' }}"
```
richm, line 87: ?

richm:
since we have to rewrite a lot of the config option get/set code, maybe we should just use the `ini_file` module https://docs.ansible.com/ansible/2.8/modules/ini_file_module.html#ini-file-module ?

it looks like the mssql.conf file format is INI?  https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-configure-mssql-conf?view=sql-server-ver15#mssql-conf-format

spetrosi:
> since we have to rewrite a lot of the config option get/set code, maybe we should just use the `ini_file` module https://docs.ansible.com/ansible/2.8/modules/ini_file_module.html#ini-file-module ?
> 
I tried to take advantage of this module, but it turned out that the module does not really provide what the role needs.
In this PR, I need to rewrite only the `mssql-conf get` code because 2017 does not support it. Instead, I grep the config file for the setting so that it returns the same output as from the `get` command.
The issue with `ini_file` is that it only can write to ini files, it cannot get settings from ini files. I would only be able to use it instead of the `mssql-conf set` command, however, set command is available in 2017 too, hence no need to do this.
I found an `ini` lookup but found it more complicated to use then grepping the file.
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/ini_lookup.html

spetrosi:
rebased to latest master

tasks/main.yml
spetrosi: Upgrade from mssql-server-14 to mssql-server-15 did not work for me without this workaround. When upgrading mssql-server with the `package` module to `state: latest`, Ansible threw dependency errors. I will look at it more closely later.

richm:
> > since we have to rewrite a lot of the config option get/set code, maybe we should just use the `ini_file` module https://docs.ansible.com/ansible/2.8/modules/ini_file_module.html#ini-file-module ?
> 
> I tried to take advantage of this module, but it turned out that the module does not really provide what the role needs.
> In this PR, I need to rewrite only the `mssql-conf get` code because 2017 does not support it. Instead, I grep the config file for the setting so that it returns the same output as from the `get` command.
> The issue with `ini_file` is that it only can write to ini files, it cannot get settings from ini files. I would only be able to use it instead of the `mssql-conf set` command, however, set command is available in 2017 too, hence no need to do this.

What is the use case for getting settings from ini files, other than to check if the value already exists?  Because that's what the ini_file module should be doing - when you use it to set a value, it should check if the value already exists

> I found an `ini` lookup but found it more complicated to use then grepping the file.
> https://docs.ansible.com/ansible/latest/collections/ansible/builtin/ini_lookup.html

tasks/main.yml
spetrosi: When upgrading using just `name: "{{ __mssql_server_packages }}"`, the role fails with the following message:

=> {"changed": false, "failures": [], "msg": "Depsolve Error occured:
Problem:
- package python36-3.6.8-2.module+el8.1.0+3334+5cb623d7.x86_64 requires python3-pip, but none of the providers can be installed
- package python3-pip-9.0.3-18.el8.noarch requires platform-python-pip = 9.0.3-18.el8, but none of the providers can be installed
- package mssql-server-15.0.4138.2-1.x86_64 requires python3, but none of the providers can be installed
- cannot install both platform-python-pip-9.0.3-18.el8.noarch and platform-python-pip-9.0.3-19.el8.noarch
- cannot install the best update candidate for package mssql-server-14.0.3391.2-12.x86_64
- problem with installed package platform-python-pip-9.0.3-19.el8.noarch",
"rc": 1, "results": []}

I think this is because Ansible is confused by the mssql-server dependencies between versions mssql-server-14 and mssql-server-15. Ansible kinda wants to update mssql-server-14 to its latest version, and at the same gets carried away by upgrading to mssql-server-15, and pulling all the dependencies there.

This issue is reproducible with yum too. When mssql-server-14 is installed and the repository for mssql-server-15 is configured, running `yum update mssql-server` results in the same error. While running `yum install mssql-server` pulls the dependencies correctly.

When I use `name: "{{ __mssql_server_packages }}-1*"` in the module, Ansible upgrades mssql-server to mssql-server-15 successfully, which eliminates the error.

This workaround is confusing, I will change this to the following explicit solution instead:
```
package:
  name: >-
    {{ __mssql_server_packages + '-15*' if mssql_upgrade
    else __mssql_server_packages }}
```

spetrosi:
> > > since we have to rewrite a lot of the config option get/set code, maybe we should just use the `ini_file` module https://docs.ansible.com/ansible/2.8/modules/ini_file_module.html#ini-file-module ?
> > 
> > I tried to take advantage of this module, but it turned out that the module does not really provide what the role needs.
> > In this PR, I need to rewrite only the `mssql-conf get` code because 2017 does not support it. Instead, I grep the config file for the setting so that it returns the same output as from the `get` command.
> > The issue with `ini_file` is that it only can write to ini files, it cannot get settings from ini files. I would only be able to use it instead of the `mssql-conf set` command, however, set command is available in 2017 too, hence no need to do this.
> 
> What is the use case for getting settings from ini files, other than to check if the value already exists? Because that's what the ini_file module should be doing - when you use it to set a value, it should check if the value already exists

The first issue lies in the way the mssql.conf file behaves. It does not store all mssql-server settings by default. After fresh install, it only has the following content:
```
[sqlagent]
enabled = false

[EULA]
accepteula = Y
```
This means that the role still needs to have some logic to check for default values. E.g. when changing the TCP port to the default 1443 on a freshly installed mssql, the role must read mssql.conf and see if the tcp_port entry exists. If it does not exist - then the port is set to 1443 already, as this is the default.

Secondly, the `mssql-conf set` does some validation, e.g. it won't let you set IP address to an invalid 1111.11111.1111123.124123, or port to an invalid 8-digit number. With ini_file module, the role would need to check such things manually, meaning that instead of 1 task of the `mssql-conf set` command, the role would have many tasks for each settings and its validation.

richm:
> > > > since we have to rewrite a lot of the config option get/set code, maybe we should just use the `ini_file` module https://docs.ansible.com/ansible/2.8/modules/ini_file_module.html#ini-file-module ?
> > > 
> > > 
> > > I tried to take advantage of this module, but it turned out that the module does not really provide what the role needs.
> > > In this PR, I need to rewrite only the `mssql-conf get` code because 2017 does not support it. Instead, I grep the config file for the setting so that it returns the same output as from the `get` command.
> > > The issue with `ini_file` is that it only can write to ini files, it cannot get settings from ini files. I would only be able to use it instead of the `mssql-conf set` command, however, set command is available in 2017 too, hence no need to do this.
> > 
> > 
> > What is the use case for getting settings from ini files, other than to check if the value already exists? Because that's what the ini_file module should be doing - when you use it to set a value, it should check if the value already exists
> 
> The first issue lies in the way the mssql.conf file behaves. It does not store all mssql-server settings by default. After fresh install, it only has the following content:
> 
> ```
> [sqlagent]
> enabled = false
> 
> [EULA]
> accepteula = Y
> ```
> 
> This means that the role still needs to have some logic to check for default values. E.g. when changing the TCP port to the default 1443 on a freshly installed mssql, the role must read mssql.conf and see if the tcp_port entry exists. If it does not exist - then the port is set to 1443 already, as this is the default.
> 
> Secondly, the `mssql-conf set` does some validation, e.g. it won't let you set IP address to an invalid 1111.11111.1111123.124123, or port to an invalid 8-digit number. With ini_file module, the role would need to check such things manually, meaning that instead of 1 task of the `mssql-conf set` command, the role would have many tasks for each settings and its validation.

Ok.  I think this is ok for now.  But if we have to make major changes to this role in the future, we should consider writing our own mssqlconf module.

tasks/main.yml
richm: The usual workflow for doing a major upgrade is:
* ensure all packages are up-to-date from the old repos
* disable the old repos
* enable the new repos
* install/update the packages from the new repos

richm:
[citest pending]

tasks/main.yml
spetrosi: Never mind, I just realized that the issue I was facing was caused by the older RHEL 8 guest image I run the tests against. I have now removed the weird workaround.

I have implemented updating all packages from the 2017 repo here:
https://github.com/linux-system-roles/mssql/pull/22/files#diff-3d0ff1709ca48add100327bb2a468e6c508fb92a159c64c4f99ad1df89d9bddeR77-R84

Is that what you expected?

tasks/main.yml
```diff
@@ -74,15 +74,24 @@
     key: https://packages.microsoft.com/keys/microsoft.asc
     state: present
 
+# This works only on systems using yum or dnf
+- name: Update all packages from SQL Server 2017 repo
+  package:
+    name: "*"
+    disablerepo: "*"
+    enablerepo: packages-microsoft-com-mssql-server-2017
+    state: latest  # noqa E403
```
richm, line 83: Looks like this isn't working to suppress the ansible-lint warning

spetrosi: Fixed with `# noqa 403`

richm (approved):
lgtm

README.md
```diff
@@ -45,7 +45,7 @@ Default: `false`
 
 Type: `bool`
 
-### `mssql_accept_microsoft_sql_server_2019_standard_eula`
+### `mssql_accept_microsoft_sql_server_standard_eula`
```
spetrosi, line 48: @richm this PR changes the variable to remove `2019` from it. The role has been released to Galaxy already, should I keep the older `mssql_accept_microsoft_sql_server_2019_standard_eula` variable as an alias?

richm: Yes, good idea.

nhosoi:
lgtm, too.

I have 2 questions about the "failures" in the upgrade. You have 2017 and upgrade it to 2019. Good. Then, you repeat the ansible-playbook with upgrade true. I'd expect ansible would successfully run with no changes as in each task... A scenario my mind is using mssql role in a larger playbook in which not just upgrading mssql itself, but also updating some other services and data. The customer may need to repeat running the playbook. In the multiple attempts, they may not want to get bothered to modify the value of mssql_upgrade and mssql_version...

Also, it's not allowed to downgrade. (yum/dnf does.:) Is it common for mssql?

richm:
> Also, it's not allowed to downgrade. (yum/dnf does.:) Is it common for mssql?

No, downgrade is not supported.

spetrosi:
> lgtm, too.
> 
> I have 2 questions about the "failures" in the upgrade. You have 2017 and upgrade it to 2019. Good. Then, you repeat the ansible-playbook with upgrade true. I'd expect ansible would successfully run with no changes as in each task... A scenario my mind is using mssql role in a larger playbook in which not just upgrading mssql itself, but also updating some other services and data. The customer may need to repeat running the playbook. In the multiple attempts, they may not want to get bothered to modify the value of mssql_upgrade and mssql_version...
You are right, I have removed this check now in The role does not support downgrading in c89556a8a4bc3629b99452bbe24feba1a7953f57.
> 
> Also, it's not allowed to downgrade. (yum/dnf does.:) Is it common for mssql?
It is not only about downgrading the package. Databases must be downgraded too, which is not the case now. I have added notes that downgrade is not supported in README and within the role itself in efff9a4d42a515205b1222cf44c434ad89b4acc3. The role will tell that downgrade is not supported when it is run with `mssql_server: 2017` against SQL Server 2019.

I have also pushed 80a807cd4a724cdd5b626e504dbad484bcc36119 that removes the requirement for users to provide `mssql_server` strictly in type int. Now, this variable can be provided  with `mssql_server: 2019` as well as `mssql_server: "2019"` and the role will eat it.

README.md
spetrosi: Done in 7f0911d62e9eaf0e8016ccf009e6c0055b4133ac

spetrosi:
[citest bad]

nhosoi (approved):
lgtm
Thank you for considering my questions, @spetrosi!

spetrosi:
[citest bad]

PR #30, opened 2021-06-30
Clarify that the role expects users to have tls cert and key

### LINKED ISSUE #27: Clarify how TLS files are created and used
It isn't clear from the documentation that the role *does not* create the TLS cert/key files if not specified.  We should state something like "You are responsible for creating and securing the TLS cert and key file.  It is assumed you have a CA which can issue these files.  If not, you can use the `openssl` command to create these files."

spetrosi:
Resolves #27


richm (approved):
lgtm - @dpredhat?

PR #31, opened 2021-06-30
Improve example playbooks to be more specific

### LINKED ISSUE #26: README should explicitly state which parameters are required and which are optional
It isn't clear from the README which parameters are required and which are optional.  The description for each parameter should specify if this parameter is required or optional.

We should provide in the README an example playbook which sets the minimal required parameters, and perhaps use the current example as an example of how to also set the optional parameters.  Then users could copy/paste the minimal and add their own as needed.
@spetrosi

### LINKED ISSUE #28: Example should use the collection role name
The example in the README should use the role name `microsoft.sql.server` - I don't think we are going to publish the legacy role to Galaxy.

spetrosi:
Resolves #28
Add example playbooks to demonstrate what variables to use in different
situations

Resolves #26
Call the role by microsoft.sql.server to comply with the collection
format

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,31 @@ Default: `false`
 
 Type: `bool`
 
-## Example Playbook
+## Example Playbooks
+
+This section outlines example playbooks that you can use as a reference.
+
+### Setting up SQL Server
+
+This example shows how to use the role to set up SQL Server with the minimum
+required variables.
+
+```yaml
+- hosts: all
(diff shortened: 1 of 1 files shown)


richm (approved):
lgtm - @dpredhat?

dpredhat:
> lgtm - @dpredhat?

Thank you, this looks great!

PR #32, opened 2021-06-30
Install openssl on the control node

spetrosi:
The role uses openssl to generate TLS certificate and key.
Previously, OpenSSL was a dependency of pcs. That is no longer the case
since pcs-0.10.8-2.
Hence, the tls tests fail on Fedora because openssl is missing.

### DIFF
--- a/tests/tests_tls.yml
+++ b/tests/tests_tls.yml
@@ -7,6 +7,12 @@
     mssql_accept_microsoft_cli_utilities_for_sql_server_eula: true
     mssql_accept_microsoft_sql_server_2019_standard_eula: true
   tasks:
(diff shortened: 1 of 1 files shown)


spetrosi:
I am not sure about this PR now because tests in #34 run just fine. I wonder if the latest CI images had the openssl package returned. I'll wait till the end of the tests in #34 and then close.

richm:
> I am not sure about this PR now because tests in #34 run just fine. I wonder if the latest CI images had the openssl package returned. I'll wait till the end of the tests in #34 and then close.

I'm not sure, but if the test uses openssl, then the right thing to do is the test should install openssl

PR #33, opened 2021-06-30
Fix failing when changing TCP port to the default 1433

### LINKED ISSUE #25: setting mssql_tcp_port to 1433 will cause an error
If you set `mssql_tcp_port: 1433` then the role will fail with an error that there is already a service listening to 1433
@spetrosi

spetrosi:
Also, add this check to the tls test
Resolves #25


richm (approved):
lgtm

PR #34, opened 2021-06-30
Move PowerShell install to a separate test

spetrosi:
Fedora VMs do not have enough disk space to install Powershell on top of
the packages installed in the tests_idempotency playbook, hence moving
the install to a separate test for now. Later, when we will use BaseOS
CI, we will be able to increase the disk size of Fedora VMs.


richm (approved):
lgtm

PR #36, opened 2021-07-01
Install openldap-compat on fedora to provide libldap-2.4.so.2

### LINKED ISSUE #35: mssql-server does not work on latest Fedora
The `mssql-server` service fails with the following error:

`Jul 01 07:48:48 fedora sqlservr[14312]: /opt/mssql/bin/sqlservr: error while loading shared libraries: libldap-2.4.so.2: cannot open shared object file: No such file or directory`

Seems like `sqlservr` has been compiled with a later `openldap` package than the one that Fedora ships.

On RHEL 8:
```
[root@localhost ~]# rpm -q openldap
openldap-2.4.46-16.el8.x86_64
```

On Fedora 34:
```
[root@fedora ~]# rpm -q openldap
openldap-2.4.57-5.fc34.x86_64
```

`mssql-server` does not officially support Fedora, it just happened to work on Fedora before. Fedora uses RHEL 8 repositories from https://packages.microsoft.com/rhel/8/mssql-server-2019/ because Fedora repositories at https://packages.microsoft.com/fedora/34/ does not provide mssql-server-2019 repository.

spetrosi:
Resolves  #35

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -44,6 +44,15 @@
     name: "{{ __mssql_server_packages }}"
     state: present
 
+# /opt/mssql/bin/sqlservr requires libldap-2.4.so.2. Latest Fedora use newer
+# libldap by default, therefore, it is required to install openldap-compat
+# because it provides libldap-2.4.so.2.
+- name: Ensure that the openldap-compat package is installed
+  package:
+    name: openldap-compat
+    state: present
+  when: ansible_distribution == "Fedora"
+
 - name: Check if the errorlog file exists and its location
   shell: |
     set -euo pipefail
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -51,7 +51,9 @@
   package:
     name: openldap-compat
     state: present
-  when: ansible_distribution == "Fedora"
+  when:
+    - ansible_distribution == "Fedora"
+    - ansible_distribution_version | int >= 34
 
 - name: Check if the errorlog file exists and its location
   shell: |


tasks/main.yml
```diff
@@ -44,6 +44,15 @@
...
     state: present
 
+# /opt/mssql/bin/sqlservr requires libldap-2.4.so.2. Latest Fedora use newer
+# libldap by default, therefore, it is required to install openldap-compat
+# because it provides libldap-2.4.so.2.
+- name: Ensure that the openldap-compat package is installed
+  package:
+    name: openldap-compat
+    state: present
+  when: ansible_distribution == "Fedora"
```
richm, line 54: I think it would be better to add a file `vars/Fedora.yml` like this:
```
__mssql_server_packages: "{{ __mssql_server_packages + ['openldap-compat'] }}"
```
I think that should work - if not, then you'll have to do it like this:
```
__mssql_server_packages: [mssql-server openldap-compat]
```

spetrosi: Tests demonstrated that the openldap-compat package is only missing on Fedora 34+, not on Fedora 33.
One way to fix this that I can think of, is to  create `vars/Fedora.yml` with the `openldap-compat` package listed, and `vars/Fedora_33.yml` without that package listed. This will fix the issue for Fedora 33 users, but not for Fedora < 33 users.
Because `vars/` allows configuring variables for separate versions only, not for e.g. Fedora <34, it makes it hard to maintain such a solution.
What you don't like about the task with conditions specifically?
I am adding a condition for ansible_distribution_version too.

richm: The issue is that this clutters up tasks/main.yml with a lot of unnecessary platform specific stuff - but ok - I can see that this is a special case - so let's go with this for now, and revisit once we drop support for Fedora 33

richm (approved):

PR #62, opened 2022-04-01
Add support for configuring HA cluster

spetrosi:
TODO:
Add mssql_ha_reset_cert
Add idempotency to all templates


tasks/main.yml
```diff
@@ -475,10 +430,194 @@
...
+      include_tasks: input_sql_file.yml
+
+    - name: Create and back up DBM certificate
+      vars:
+        __mssql_input_sql_file: create_and_back_up_dbm_cert.j2
+      include_tasks: input_sql_file.yml
+
+    - name: Fetch DBM certificates from the primary node to the control node
+      fetch:
+        src: "{{ item.value}}"
```
richm, line 477: add space before `}}`

tasks/main.yml
```diff
@@ -475,10 +430,194 @@
...
+      with_dict:
+        cert: /var/opt/mssql/data/{{ mssql_ha_dbm_cert_name }}.cer
+        key: /var/opt/mssql/data/{{ mssql_ha_dbm_cert_name }}.pvk
+
+    - name: Create database mirroring endpoints
+      vars:
+        __mssql_input_sql_file: create_endpoint.j2
+      include_tasks: input_sql_file.yml
+
+    - name: Get mssql-server version to see if WRITE_LEASE_VALIDITY is available
```
richm, line 489: The task name doesn't seem to have anything to do with the task action?

tasks/main.yml
```diff
@@ -475,10 +430,194 @@
         __mssql_conf_setting: "network forceencryption"
         __mssql_conf_setting_value: "{{ '1' if mssql_tls_enable else 'unset' }}"
 
+- name: Configure availability group on the primary node
+  when:
+    - mssql_ha_configure is defined
+    - mssql_ha_configure | bool
+    - mssql_ha_replica_type == 'primary'
```
richm, line 437: so in the inventory for this host you would have a host var `mssql_ha_replica_type: primary`, and for the replica hosts, you would have host vars for `mssql_ha_replica_type: replica` ?

templates/create_ag.j2
```diff
@@ -0,0 +1,43 @@
...
+IF EXISTS (SELECT * FROM sys.availability_groups WHERE name = '{{ mssql_ha_ag_name }}')
+BEGIN
+  PRINT 'Availability group {{ mssql_ha_ag_name }} already exists, removing';
+  DROP AVAILABILITY GROUP {{ mssql_ha_ag_name }};
+  PRINT 'Availability group {{ mssql_ha_ag_name }} removed successfully';
+END
+
+PRINT 'Creating the {{ mssql_ha_ag_name }} availability group';
+CREATE AVAILABILITY GROUP {{ mssql_ha_ag_name }}
+{% if ansible_os_family == 'RedHat' and ansible_distribution_version | float < 8.3 %}
```
richm, line 11: Use the `version` test instead of `float` - https://docs.ansible.com/ansible/2.9/user_guide/playbooks_tests.html#version-comparison

tests/tests_configure_ha_cluster.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Open required firewall ports
+      vars:
+        firewall:
+          - port: "{{ mssql_ha_listener_port }}/tcp"
+            zone: public
+            state: enabled
+            permanent: true
+            runtime: true
+      include_role:
+        name: fedora.linux_system_roles.firewall
```
richm, line 24: hmm - I guess this is ok for now - for CI you'll need to add a meta/requirements.yml with the `fedora.linux_system_roles` collection 
```
collections:
  - name: fedora.linux_system_roles
```
tox -e qemu[-ansible-core-x.xx] will also use this file to install necessary collections

tasks/main.yml
```diff
@@ -49,6 +54,32 @@
...
+- name: Verify that 'mssql_ha_replica_type = primary' is provided once
+  assert:
+    that: |-
+      {% set result = namespace(total=0) %}
+      {% for item in ansible_play_hosts_all %}
+      {% if hostvars[item]['mssql_ha_replica_type'] == 'primary' %}
+      {% set result.total = 1 + result.total %}
+      {% endif %}
+      {% endfor %}
+      {{ result.total }} == 1
```
richm, line 76: In almost every case, doing something like `x = x + 1` in a loop in ansible or jinja is an anti-pattern.  I think you can do something like this instead:
```yaml
- name: Verify that 'mssql_ha_replica_type = primary' is provided once
  assert:
    that: ansible_play_hosts_all | map('extract', hostvars, 'mssql_ha_replica_type') | select('match', '^primary$') | length == 1
```
That is, ensure that at least one, and only one, host has `mssql_ha_replica_type=primary`

tasks/main.yml
```diff
@@ -475,10 +457,227 @@
...
+        state: "{{ 'present' if mssql_ha_configure | bool else 'absent' }}"
+      register: __mssql_server_ha_packages_install
+
+    - name: Enable the hadrenabled setting
+      include_tasks: mssql_conf_setting.yml
+      vars:
+        __mssql_conf_setting: "hadr hadrenabled"
+        __mssql_conf_setting_value: 1
+
+    # meta: flush_handlers does not support when conditional
```
richm, line 479: right - and in this case I think the explicit restart is the right thing to do

tasks/main.yml
```diff
@@ -475,10 +457,227 @@
...
+      vars:
+        __mssql_input_sql_file: replicate_db.j2
+      include_tasks: input_sql_file.yml
+
+    # This is required because `any_errors_fatal: true` does not work within
+    # blocks that have rescue or always sections
+    - name: Set a fact to indicate successfull set up on the primary replica
+      delegate_to: localhost
+      set_fact:
+        __mssql_primary_successfull: true
```
richm, line 548: successful - I know, English is a weird language :P

tasks/main.yml
```diff
@@ -3,6 +3,7 @@
 - name: Set platform/version specific variables
   include_vars: "{{ __mssql_vars_file }}"
   loop:
+    - main.yml
```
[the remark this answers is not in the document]
richm: > I was facing an issue where a new variable in main.yml was not included in hostvars, so did this as a solution. I'll see what caused it, maybe hostvars were cached and were not populated with new variables.

That's really strange - does running the playbook with `-vvvvv` give additional information?  We may need to get someone from the ansible team to help with this.  afaik, you should never have to explicitly load vars from vars/main.yml - it should happen automatically when the role is loaded.

tasks/main.yml
spetrosi: I've added a comment above to clarify that write_lease_validity is used in configure_ag.j2, there it is added based on the RHEL version.

templates/create_ag.j2
spetrosi: Done, didn't know this thank you

tasks/main.yml
spetrosi: Done, thank you, your solution works

tests/tests_configure_ha_cluster.yml
spetrosi: I'm planing to take a look at the CI and how to implement it a bit later

spetrosi:
[citest pending]

spetrosi:
@richm `rhel-8-y/ansible-2.12` and `rhel-8/ansible-2.12` CI tests fail with `include_role: name: fedora.linux_system_roles.ha_cluster` being incorrect. Although other tests are fine with this? Do you have any idea why this can be?

spetrosi:
[citest pending]

tests/tests_configure_ha_cluster.yml
```diff
@@ -0,0 +1,81 @@
...
+    #         permanent: true
+    #         runtime: true
+    #     when: >-
+    #       ('running' in
+    #       ansible_facts['services']['firewalld.service']['state']) or
+    #       ('enabled' in
+    #       ansible_facts['services']['firewalld.service']['status'])
+    #   include_role:
+    #     name: fedora.linux_system_roles.firewall
+    - name: Skip this test on EL < 7 because ha_cluster does not support it
```
nhosoi, line 33: A very minor question... Why `EL < 7` here, instead of `EL < 8.3`?

README.md
```diff
@@ -417,6 +566,56 @@ use TLS encryption.
...
+- hosts: all
+  vars:
+    mssql_accept_microsoft_odbc_driver_17_for_sql_server_eula: true
+    mssql_accept_microsoft_cli_utilities_for_sql_server_eula: true
+    mssql_accept_microsoft_sql_server_standard_eula: true
+    mssql_password: "p@55w0rD"
+    mssql_edition: Evaluation
+    mssql_ha_configure: true
+    mssql_ha_firewall_configure: true
+    mssql_ha_listener_port: 5022
```
nhosoi, line 602: Another minor question... If you plan to use the default values, you don't have to declare the variables `mssql_ha_firewall_configure` and `mssql_ha_listener_port`, right? Of course, it's ok to set them, though... This question is applied to tests_configure_ha_cluster.yml, as well. Maybe, it's a good practice type of issue...? :) Setting all the available variables vs. setting as little variables as possible.

README.md
```diff
@@ -327,6 +327,155 @@ Default: `{{ __mssql_client_repository }}`
...
+
+See [`Setting Up SQL Server and Configuring for High Availability`](#Setting-Up-SQL-Server-and-Configuring-for-High-Availability) for an example inventory.
+
+You must set the `mssql_ha_replica_type` variable to `primary` for exactly one host.
+
+The available values are: `primary`, `synchronous`, `witness`.
+
+Default: no
+
+Type: `bool`
```
richm, line 364: Type: `str`

tasks/input_sql_file.yml
```diff
@@ -0,0 +1,77 @@
...
+      changed_when: '"successfully" in __mssql_sqlcmd_input_file.stdout'
+
+  always:
+    - name: >-
+        Print the output of the sqlcmd command for {{ __mssql_input_sql_file }}
+        if not empty
+      debug:
+        var: __mssql_sqlcmd_input_file.stdout_lines
+      changed_when: false
+      when: __mssql_sqlcmd_input_file.stdout_lines is defined
```
richm, line 65: Is it possible that this output could contain cleartext passwords or other plaintext secrets?
If the user might want to turn off this output, add a condition using `__mssql_debug`
```
when:
  - __mssql_debug | d(false)
  - __mssql_sqlcmd_input_file.stdout_lines is defined
```

templates/create_and_back_up_cert.j2
```diff
@@ -0,0 +1,73 @@
+-- Enabling NOCOUNT to suppress (1 rows affected) messages from DECLARE
+-- keywordss on the output
```
richm, line 2: keywords

tests/tests_configure_ha_cluster.yml
```diff
@@ -0,0 +1,81 @@
...
+      set_fact:
+        mssql_ha_replica_type: primary
+      when: ansible_play_hosts_all |
+            map('extract', hostvars, 'mssql_ha_replica_type') |
+            list is not defined
+
+    - name: Set up test environment for the ha_cluster role
+      include_role:
+        name: fedora.linux_system_roles.ha_cluster
+        tasks_from: ../../../tests/ha_cluster/tasks/setup_test.yml
```
richm, line 49: I don't like using relative paths like this - what if we change the ha_cluster role to have a `ha_cluster_setup_for_test: true` and `ha_cluster_setup_sbd_for_test: true` parameters?  Then you could use
```yaml
    - name: Set up test environment and SBD for the ha_cluster role
      include_role:
        name: fedora.linux_system_roles.ha_cluster
        vars:
          ha_cluster_setup_for_test: true
          ha_cluster_setup_sbd_for_test: true
```
What you have is OK for now, but we should change this in the future.

richm:
[citest bad]

README.md
spetrosi: Fixed

tests/tests_configure_ha_cluster.yml
spetrosi: I've implemented your suggestion in https://github.com/linux-system-roles/ha_cluster/pull/50

README.md
```diff
@@ -327,6 +338,164 @@ Default: `{{ __mssql_client_repository }}`
 
 Type: `string`
 
+### `mssql_ha_configure`
+
+Use the variables with the `mssql_ha_` prefix to configure SQL Server for high
```
limershein, line 343: Use the variable with the 'mssql_ha_' prefix to configure a SQL Server Always On availability group to provide high availability.

README.md
```diff
@@ -327,6 +338,164 @@ Default: `{{ __mssql_client_repository }}`
...
+
+The available values are: `primary`, `synchronous`, `witness`.
+
+Default: no
+
+Type: `str`
+
+#### `mssql_ha_firewall_configure`
+
+Whether to configure firewall for high availability or not.
```
limershein, line 383: Whether to open ports in the Linux firewall for the Always On availability group.  This works for firewalld, if you are using different firewall software, you will need to open the ports through different means.

README.md
```diff
@@ -327,6 +338,164 @@ Default: `{{ __mssql_client_repository }}`
...
+The role configures firewall only if the `firewalld` service has been started or
+enabled.
+
+Default: `true`
+
+Type: `bool`
+
+#### `mssql_ha_listener_port`
+
+The TCP port to be used for high availability.
```
limershein, line 394: The TCP port used to replicate data for an Always On availability group.

README.md
```diff
@@ -327,6 +338,164 @@ Default: `{{ __mssql_client_repository }}`
...
+
+Default: `5022`
+
+Type: `int`
+
+#### `mssql_ha_cert_name`
+
+The name of the certificate used for high availability configuration.
```
limershein, line 402: The name of the certificate used to secure transactions between members of an Always On availability group.

README.md
```diff
@@ -327,6 +338,164 @@ Default: `{{ __mssql_client_repository }}`
...
+
+The password to set for the private key used with the certificate.
+
+Default: `null`
+
+Type: `str`
+
+#### `mssql_ha_reset_cert`
+
+Whether to reset certificates used for high availability or not.
```
limershein, line 426: Whether to reset certificates used by the Always On availability group or not.

README.md
```diff
@@ -327,6 +338,164 @@ Default: `{{ __mssql_client_repository }}`
...
+The name of the database to be replicated. This database must exist in SQL
+Server.
+
+Default: `null`
+
+Type: `string`
+
+#### `mssql_ha_db_backup_path`
+
+The path to back up replicated database to.
```
limershein, line 460: For SQL Server, any database participating in an Availability Group must be in full recovery mode and  have a valid log backup. This path will be used to backup the database prior to initiating replication within the Always On availability group.

README.md
```diff
@@ -327,6 +338,164 @@ Default: `{{ __mssql_client_repository }}`
...
+
+Default: `/var/opt/mssql/data/{{ mssql_ha_db_name }}.bak`
+
+Type: `string`
+
+#### `mssql_ha_login`
+
+The user to be created for pacemaker in SQL Server.
```
limershein, line 468: The user created for pacemaker in SQL Server.  This user will be used by the SQL Server Pacemaker resource agent to connect to SQL Server to perform regular database health checks and manage state transitions from replica to primary when needed.

README.md
```diff
@@ -327,6 +338,164 @@ Default: `{{ __mssql_client_repository }}`
...
+
+The password for the cluster to be created in pacemaker.
+
+Default: `null`
+
+Type: `string`
+
+#### `mssql_ha_virtual_ip`
+
+A floating virtual IP address for accessing the master SQL Server node to be
```
limershein, line 492: A floating virtual IP address for accessing the primary SQL Server node in an Always On availability group to be created in Pacemaker.  In the event of a node failure, a synchronous replica is automatically promoted to primary and the address is assigned to this new node.

README.md
richm: Or say that the role uses the System Roles firewall role to manage the firewall, and only the implementations supported by that role will work.

README.md
spetrosi: The role sets a full recovery mode and backs the database provided with `mssql_ha_db_backup_path` if no back-up newer than 3 hours exists. I need to mention this too in README for transparency.

README.md
spetrosi: Thank you, applied

README.md
spetrosi: Applied, thank you

README.md
spetrosi: Applied

README.md
spetrosi: Applied

README.md
spetrosi: Applied

README.md
spetrosi: Applied

README.md
spetrosi: Applied

README.md
spetrosi: Applied

spetrosi:
I'll also need to add comments to new tasks files to describe why these files are in `tasks/`.

tomjelinek (changes requested):

README.md
```diff
@@ -327,6 +338,185 @@ Default: `{{ __mssql_client_repository }}`
...
+1.1. Open the firewall port set with the `mssql_ha_listener_port` variable.
+1.2. Enable the `high-availability` service in firewall.
+2. Configure SQL Server for high availability:
+2.1. Enable AlwaysOn Health events.
+2.2. Create certificate on the primary replica and distribute to other replicas.
+2.3. Configure endpoint and availability group.
+2.4. Configure the user provided with the ``mssql_ha_login` variable for
+     Pacemaker.
+3. Include the System Roles ha_cluster role to configure pacemaker.
+   Note that this role does not configure STONITH devices in Pacemaker.
```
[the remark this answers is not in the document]
spetrosi: My intent here was to avoid making the mssql role wrap around the ha_cluster role much. We are going to provide example playbooks that will run the mssql role and then run the ha_cluster role with all requried ha_cluster-related variables.
I thought that setting `ha_cluster_resource_primitives` is not required because `stonith-enabled` is set to `true` by default, and it seemed to work. Could you please provide an example of what `mssql_ha_stonith_resources` value should look like?
In the mssql role, I guess adding `mssql_ha_stonith_resources` would look like this?
```
        ha_cluster_resource_primitives:
          - "{{ mssql_ha_stonith_resources }}"
          - id: ag_cluster
          ...
```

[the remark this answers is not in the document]
tomjelinek: > We are going to provide example playbooks that will run the mssql role and then run the ha_cluster role with all requried ha_cluster-related variables.

Ok, that would work for me. Just bear in mind this part of the ha_cluster role readme: "The role replaces the configuration of HA Cluster on specified nodes. Any settings not specified in the role variables will be lost."

> I thought that setting `ha_cluster_resource_primitives` is not required because `stonith-enabled` is set to `true` by default

That merely enables stonith. You still need to tell the cluster which stonith devices it can use, how it can connect to them, and which nodes they can fence. That is done by creating stonith resources. For further details, see for example:
* https://clusterlabs.org/pacemaker/doc/2.1/Clusters_from_Scratch/html/fencing.html
* https://clusterlabs.org/pacemaker/doc/2.1/Pacemaker_Explained/html/fencing.html
* https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_high_availability_clusters/assembly_configuring-fencing-configuring-and-managing-high-availability-clusters

> Could you please provide an example of what mssql_ha_stonith_resources value should look like?
```yaml
ha_cluster_resource_primitives:
  - id: myapc
    agent: stonith:fence_apc_snmp
    instance_attrs:
      - attrs:
          - name: ipaddr
            value: apc-switch.example.com
          - name: pcmk_host_map
            value: rhel8-node1.example.com:1;rhel8-node2.example.com:2
          - name: login
            value: apc
          - name: passwd
            value: apc
```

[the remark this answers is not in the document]
spetrosi: > Ok, that would work for me. Just bear in mind this part of the ha_cluster role readme: "The role replaces the configuration > of HA Cluster on specified nodes. Any settings not specified in the role variables will be lost."

Oh, so in the case where a user runs the mssql role and then runs the ha_cluster role with `ha_cluster_resource_primitives` containing one item, the configuration applied by the mssql role would be removed? In this case, the mssql role would need to completely wrap around the ha_cluster role. As well as any other role that would use ha_cluster.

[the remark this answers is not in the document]
spetrosi: @tomjelinek, in the case where a user runs the `mssql` role and then runs the `ha_cluster` role with `ha_cluster_resource_primitives` containing one item, the cluster configuration applied previously by the `mssql` role would be removed?

spetrosi:
[citest]

README.md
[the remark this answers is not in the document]
tomjelinek: > @tomjelinek, in the case where a user runs the `mssql` role and then runs the `ha_cluster` role with `ha_cluster_resource_primitives` containing one item, the cluster configuration applied previously by the `mssql` role would be removed?

In the case where a user runs the `mssql` role and then runs the `ha_cluster` role (with any settings), the cluster configuration applied previously by the `mssql` role would be removed.

It is unfortunate, but it is the best I can do now. Making the `ha_cluster` role work in a way where users can specify only partial cluster configuration and the role would not touch other parts of the cluster requires a lot more resources.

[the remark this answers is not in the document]
richm: > It is unfortunate, but it is the best I can do now. Making the ha_cluster role work in a way where users can specify only partial cluster configuration and the role would not touch other parts of the cluster requires a lot more resources.

I think this is fine.  We just document that you cannot use the ha_cluster role on mssql nodes, and vice versa.

README.md
```diff
@@ -512,15 +518,56 @@ Default: `null`
...
+role by setting `mssql_ha_cluster_run_role: true`.
+
+If you want to run the `ha_cluster` role independently of the
+`microsoft.sql.server` role, or run provide custom variables for the
+`ha_cluster` role, you can work around this limitation. To do this,
+run the `microsoft.sql.server` role with
+[`mssql_ha_cluster_print_vars:`](#mssql_ha_cluster_print_vars)`true` to print
+planned `ha_cluster` variables. Then you can merge the printed variables with
+your custom `ha_cluster` variables and specify the resulting set of variables
+with the `microsoft.sql.server` role invocation.
```
richm, line 540: What is the use case for this?  If you are using the ha_cluster role and the mssql role in the same playbook, and you are configuring mssql for ha, and you want to ensure you use the same ha parameters for both ha_cluster and mssql?  I guess this is if you first run mssql and configure it for ha, then do a subsequent playbook run with ha_cluster (with or without mssql) and you want to ensure that you do not overwrite the previous ha cluster settings used when deploying mssql?

richm: Or vice versa - you deploy ha_cluster role first, then you want to grab those ha_cluster variables to use with the mssql role?  If that's the case, then it seems like we need support in the ha_cluster role - have it export the settings in a format that you can use to configure something else to use ha_cluster on those same machines.  Other roles have a similar feature - the ability to export the role configuration - https://github.com/linux-system-roles/kernel_settings/pull/58 - https://github.com/linux-system-roles/firewall/pull/83 (these are wip, but I believe some other roles have already implemented this)

tasks/main.yml
```diff
@@ -37,9 +41,79 @@
     - mssql_upgrade | bool
     - mssql_version | int == 2017
 
+- name: Verify that EL < 8 is not used with mssql_ha_configure=true
+  assert:
+    that:
+      - ansible_distribution in ['CentOS', 'RedHat']
+      - ansible_distribution_version is version('8', '>')
```
nhosoi, line 48: To include EL 8, should it be `version('8', '>=')`?

tasks/main.yml
```diff
@@ -37,9 +41,79 @@
...
+    fail_msg: >-
+      mssql_ha_configure=true does not support running against EL 7 hosts
```
nhosoi, line 50: A very minor issue :)... Not "CentOS or RedHat newer than 8" includes not only `EL 7`, but also EL 6, Fedora, and other Linux distribution. I'd think they are also not supported, right? I thought it might be a bit confusing if the playbook failed when it's run against, e.g., Fedora and get `mssql_ha_configure=true does not support running against EL 7 hosts`.

tasks/main.yml
richm: maybe use `fail` here:
```yaml
- name: Verify that EL < 8 is not used with mssql_ha_configure=true
  fail:
    msg: mssql_ha_configure=true does not support running against EL 7 hosts
  when:
    - ansible_distribution in ['CentOS', 'RedHat']
    - ansible_distribution_version is version('8', '<')
```

spetrosi:
[citest pending]

PR #67, opened 2022-06-10
Add mssql_tls_remote_src

spetrosi:
SQL Server role should be able to use TLS cert and key files that are
already on managed nodes

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -420,6 +420,7 @@
     - name: Copy certificate and private_key files to the host
       copy:
         src: "{{ item }}"
+        remote_src: "{{ mssql_tls_remote_src }}"
         dest: >-
           /etc/pki/tls/{{ 'certs' if item == mssql_tls_cert
           else 'private' }}/{{ item | basename }}
--- a/tests/tests_tls_2019.yml
+++ b/tests/tests_tls_2019.yml
@@ -6,11 +6,15 @@
     mssql_accept_microsoft_odbc_driver_17_for_sql_server_eula: true
     mssql_accept_microsoft_cli_utilities_for_sql_server_eula: true
     mssql_accept_microsoft_sql_server_standard_eula: true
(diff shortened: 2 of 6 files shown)


spetrosi:
[citest pending]

spetrosi:
[citest pending]

tests/tests_tls_2019.yml
```diff
@@ -80,5 +81,39 @@
...
+    - name: Remove certificates from hosts
+      file:
+        path: /etc/pki/tls/{{ item }}
+        state: absent
+      loop: "{{ __mssql_certs }}"
+
+    - name: Copy certificates to hosts
+      copy:
+        src: "{{ item }}"
+        dest: ~/{{ item }}
```
richm, line 94: using `~` in this context works?

tasks/main.yml
```diff
@@ -420,6 +420,7 @@
     - name: Copy certificate and private_key files to the host
       copy:
         src: "{{ item }}"
+        remote_src: "{{ mssql_tls_remote_src }}"
```
richm, line 423: So this assumes the cert and key are already on the remote host, just not in the right location?

spetrosi: Yes, e.g. when using the certificate role to generate certificates and then must move them to the location required for mssql

tests/tests_tls_2019.yml
spetrosi: It did when I run CI locally with tox

richm:
[citest]

tests/tests_tls_2019.yml
spetrosi: I should rather copy them to /tmp/{{ item }} though

spetrosi:
[citest pending]

spetrosi:
[citest]

richm (approved):
lgtm

PR #69, opened 2022-06-14
Do not install mssql-server-fts

spetrosi:
This RPM is installed only once during tests and it exceeds storage


richm:
I updated the legacy CI to use a 2GB size for the yum cache, and it seems the mssql role works now.  So I would say this PR is optional - if you want to keep it, ok, but it shouldn't be necessary anymore.

spetrosi:
Thanks Rich!

PR #70, opened 2022-06-22
Refractor for ha

spetrosi:
This should help to reduce the size of #62


richm:
Are there any README.md diffs that are content related and not formatting related?

spetrosi:
> Are there any README.md diffs that are content related and not formatting related?

No, only formatting

spetrosi:
[citest bad]

PR #72, opened 2022-07-08
Use variable instead of hardcoded config file path

### LINKED ISSUE #56: use variable instead of hardcoded config file path
There are several places in the code that use the hardcoded path `/var/opt/mssql/mssql.conf` - instead, the role should declare this in `vars/main.yml`
```yaml
__mssql_conf_file: /var/opt/mssql/mssql.conf
```
and use the variable in the role

spetrosi:
fixes #56


richm (approved):
lgtm

spetrosi:
[citest]

PR #77, opened 2022-07-15
Use firewall role to configure firewall for SQL Server

spetrosi:
- Add `mssql_firewall_configure` and `mssql_firewall_close_previous`
- Set defaults for variables that decide whether to run the firewall
role to `false` so that user manually sets them to `true` because
running the firewall role enables firewall on system even if it was
disabled before
- Add an automation test for this functionality


tasks/main.yml
```diff
@@ -328,13 +328,55 @@
     __mssql_conf_setting_value: "{{ mssql_ip_address }}"
   when: mssql_ip_address is not none
 
-- name: Configure the TCP port setting
+- name: Register the previous tcpport setting
+  when:
+    - mssql_firewall_configure | bool
+    - mssql_firewall_close_previous | bool
+  shell: grep 'tcpport' {{ __mssql_conf_path }} || true
```
richm, line 335: I thought there was a mssql-conf command to do this?  If not, then you might want to be more restrictive about the pattern used e.g. `grep '^tcpport[ ]' or something like that to avoid spurious matches.  Also, using `shell` isn't recommended - is there a way to use `command` and `failed_when` instead?

tasks/main.yml
```diff
@@ -328,13 +328,55 @@
...
+  changed_when: false
+  register: __mssql_previous_tcp_port
+
+- name: >-
+    Open the {{ '1433' if mssql_tcp_port is none else mssql_tcp_port }} TCP port
```
richm, line 340: It's probably worth defining the port as a variable rather than having to repeat the expression `{{ '1433' if mssql_tcp_port is none else mssql_tcp_port }}` in several places

tasks/main.yml
```diff
@@ -328,13 +328,55 @@
...
     __mssql_conf_setting_value: "{{ mssql_tcp_port }}"
   when: mssql_tcp_port is not none
 
+- name: Close the previously set SQL Server TCP port if it changed
+  vars:
+    __tcpport: >-
+      {{ __mssql_previous_tcp_port.stdout | regex_search( '[1-9][0-9]{0,4}') }}
+    firewall:
+      - port: "{{ __tcpport if __tcpport else '1433' }}/tcp"
+        zone: public
```
richm, line 367: do you need to set the `zone` here?  If you omit the zone, does it use whatever is the default zone?  The user can then use the firewall role to set the default zone https://github.com/linux-system-roles/firewall/blob/master/tasks/main.yml#L43 (hmm - this isn't documented in the README, but it is definitely supported)

tasks/main.yml
spetrosi: It did change the behavior, I fixed so that it works everywhere.
One feature of sqlcmd is that sqlcmd does not update mssq.conf if you set tcp port to the default value 1433. It will keep the config clear.
If you set the port to some custom value 1435 and then back to 1433, the config will have `tcpport = 1433` in it.

README.md
```diff
@@ -116,6 +114,28 @@ Default: `null`
...
+
+Type: `bool`
+
+#### `mssql_firewall_close_previous`
+
+Whether to close the firewall port that was previously configured in SQL Server in case you change from one port to another when providing the `mssql_firewall_configure` variable.
+
+Default: `false`
+
+Type: `bool`
```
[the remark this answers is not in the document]
spetrosi: I introduced this for users who need this port opened, e.g. if they use it for something else. I don't think people generally use the same port for multiple services, but it can be possible.
Also, this way, user accepts that some port will be closed. 
What do you think?

tasks/main.yml
spetrosi: Omitting the zone works fine.
With zone set to public, the firewall role does this: 
```
TASK [fedora.linux_system_roles.firewall : Configure firewall] *******************************************************************************************************************************
changed: [primary] => (item={'port': '1433/tcp', 'zone': 'public', 'state': 'enabled', 'permanent': True, 'runtime': True}) => {"__firewall_changed": true, "ansible_loop_var": "item", "changed": true, "item": {"permanent": true, "port": "1433/tcp", "runtime": true, "state": "enabled", "zone": "public"}}
```
And with an unset zone this:

```
TASK [fedora.linux_system_roles.firewall : Configure firewall] *******************************************************************************************************************************
changed: [primary] => (item={'port': '1433/tcp', 'state': 'enabled', 'permanent': True, 'runtime': True}) => {"__firewall_changed": true, "ansible_loop_var": "item", "changed": true, "item": {"permanent": true, "port": "1433/tcp", "runtime": true, "state": "enabled"}}
```
So the default zone is used in this case, this should be good.

tasks/main.yml
richm: > It did change the behavior, I fixed so that it works everywhere. One feature of sqlcmd is that sqlcmd does not update mssq.conf if you set tcp port to the default value 1433. It will keep the config clear. If you set the port to some custom value 1435 and then back to 1433, the config will have `tcpport = 1433` in it.

ok, sounds good

README.md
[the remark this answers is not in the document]
spetrosi: Okay I agree, adding it was an overkill. Removed `mssql_firewall_close_previous`.

spetrosi:
[citest pending]

tasks/main.yml
spetrosi: Hi Rich,
I replaced `|| true` with `failed_when: false` and replaced `shell` with `command` because piping is now not necessary.
I made patters to be more restrictive like `'^tcpport = '` in all places.
I didn't use `/opt/mssql/bin/mssql-conf get` because it returns `No setting for the given option found in '/var/opt/mssql/mssql.conf'.` if tcpport setting is not set, even though the value is 1433 by default. And this happens inconsistently with different settings. And grepping the config file is simpler and cleaner.
I am planning to create a module for configuring mssql, it should resolve this issue in future. For now I think it's ok.

spetrosi:
[citest pending]

richm (approved):

spetrosi:
[citest pending]

PR #81, opened 2022-08-01
Close previous listener port when mssql_ha_firewall_configure: true

spetrosi:


spetrosi:
[citest pending]

spetrosi:
[citest pending]

spetrosi:
When working on this functionality, a bumped into an SQL Server limitation - it is not supported to change the LISTENER PORT of a configuration-only replica.
I tried running the following query on the primary replica to alter AG to update the port number for a configuration-only replica:
```
      ALTER AVAILABILITY GROUP ag1 MODIFY REPLICA ON
        N'kvm-02-guest02' WITH (
          ENDPOINT_URL = N'tcp://[kvm-02-guest02.rhts.eng.brq.redhat.com:5033](http://kvm-02-guest02.rhts.eng.brq.redhat.com:5033/)'
        );
```
This resulted in this error:
```
Msg 47112, Level 16, State 1, Server kvm-01-guest25, Line 89
Only the SESSION_TIMEOUT property can be modified for a configuration-only replica.
```
I confirmed with MS folks and such functionality is not supported. When users need to change a port, they need to re-create the AG.

MS folks also told me that this is not a common scenario, and they didn't have customers who needed such functionality. #91 would add a note about this to the README.

PR #82, opened 2022-08-01
tests_tls: generate certs on managed nodes to avoid installing local RPM

spetrosi:


spetrosi:
[citest pending]

spetrosi:
BaseOS CI does not have openssl and it's generally cleaner to not install additional RPMs on control node when possible.

PR #84, opened 2022-08-03
Replace simple `mssql_input_sql_file` with `pre` and `post` variables

spetrosi:
Previously, `mssql_input_sql_file` input SQL files at the end of the
role invocation. However, users must input some SQL files at the
beginning of the role before it applies any configuration to SQL Server.
`mssql_pre_input_sql_file` and `mssql_post_input_sql_file` has been
added to address this functionality.

PR #88, opened 2022-08-09
Remove unnecessary quotes from yaml strings

### LINKED ISSUE #86: Remove unnecessary quotes from yaml strings
If you are using the yaml string > or | (with and without the - modifier) you do not have to quote the string.
E.g. this:
```
    - name: Print that the variable is deprecated
      debug:
        msg: >-
          "The accept_microsoft_sql_server_2019_standard_eula variable is
          deprecated and will be removed in a future version. Edit your playbook
          to use the mssql_accept_microsoft_sql_server_standard_eula variable
          instead."
```
Must be this:

```
    - name: Print that the variable is deprecated
      debug:
        msg: >-
          The accept_microsoft_sql_server_2019_standard_eula variable is
          deprecated and will be removed in a future version. Edit your playbook
          to use the mssql_accept_microsoft_sql_server_standard_eula variable
          instead.
```

spetrosi:
Fixes #86

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -19,10 +19,10 @@
     - name: Print that the variable is deprecated
       debug:
         msg: >-
-          "The accept_microsoft_sql_server_2019_standard_eula variable is
+          The accept_microsoft_sql_server_2019_standard_eula variable is
           deprecated and will be removed in a future version. Edit your playbook
           to use the mssql_accept_microsoft_sql_server_standard_eula variable
-          instead."
+          instead.
 
     - name: >-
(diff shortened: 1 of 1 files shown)


spetrosi:
@richm so two places where I introduced redundant quotes in #84 are the only places where such quotes occur... Other places have syntax like this, which will break without the quotes:
```
        - name: Assert expected messages
          assert:
            that: >-
              "Certificate {{ mssql_ha_cert_name }} created successfully"
              in __mssql_sqlcmd_input_file.stdout

```
I could have just fixed this in #84, but at least I checked :)

richm (approved):

PR #91, opened 2022-08-18
README: Note that changing HA listener port number is not supported

spetrosi:


spetrosi:
This is a follow-up of #81.

richm (approved):

PR #98, opened 2022-08-22
Sort ansible_play_hosts to ensure configure_ag script runs smoothly

spetrosi:
Sort primary replica as the last in the list
Sort witness replica as the next-to-last in the list
Configure primary last because it has information about other replicas
Configure witness next-to-last to ensure that previous witness replica is
removed beforehand. Otherwise, script fails with 'only supports one replica

### DIFF
--- a/templates/configure_ag.j2
+++ b/templates/configure_ag.j2
@@ -33,7 +33,24 @@ BEGIN
     PRINT 'DB_FAILOVER = ON is already set, skipping'
   END
   PRINT 'Verifying replicas'
+{# Sort ansible_play_hosts #}
+{# Sort primary replica as the last in the list #}
+{# Sort witness replica as the next-to-last in the list #}
+{# Configure primary last because it has information about other replicas #}
+{# Configure witness next-to-last to ensure that previous witness replica is #}
+{# removed beforehand. Otherwise, the script fails with 'only supports one #}
+{# replica which has configuration-only availability mode.' #}
+{% set ag_replicas_sorted = [] %}
 {% for item in ansible_play_hosts %}
+{%   if hostvars[item]['mssql_ha_replica_type'] == 'primary' %}
+{{ ag_replicas_sorted.insert(-1, item) -}}
+{%   elif hostvars[item]['mssql_ha_replica_type'] == 'witness' %}
+{{ ag_replicas_sorted.insert(-2, item) -}}
+{%   else  %}
+{{ ag_replicas_sorted.insert(0, item) -}}
+{%   endif %}
+{% endfor %}
+{% for item in ag_replicas_sorted %}
 {%   if hostvars[item]['mssql_ha_replica_type'] != 'absent' %}
   IF EXISTS (
     SELECT replica_server_name, availability_mode_desc
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -626,7 +626,7 @@
         key: "{{ __mssql_ha_private_key_dest }}"
       changed_when: false
 
-    - name: Create database mirroring endpoints
+    - name: Configure database mirroring endpoints
       vars:
         __mssql_input_sql_file: configure_endpoint.j2
       include_tasks: input_sql_file.yml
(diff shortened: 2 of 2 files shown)


templates/configure_ag.j2
```diff
@@ -33,7 +33,24 @@ BEGIN
...
 {% for item in ansible_play_hosts %}
+{%   if hostvars[item]['mssql_ha_replica_type'] == 'primary' %}
+{{ ag_replicas_sorted.insert(-1, item) -}}
+{%   elif hostvars[item]['mssql_ha_replica_type'] == 'witness' %}
+{{ ag_replicas_sorted.insert(-2, item) -}}
+{%   else  %}
+{{ ag_replicas_sorted.insert(0, item) -}}
+{%   endif %}
+{% endfor %}
+{% for item in ag_replicas_sorted %}
```
richm, line 53: I don't think this will work:
```
>python
Python 3.10.6 (main, Aug  2 2022, 00:00:00) [GCC 12.1.1 20220507 (Red Hat 12.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> ary = []
>>> ary.insert(0, "normal")
>>> ary
['normal']
>>> ary.insert(-1, "primary")
>>> ary
['primary', 'normal']
>>> ary.insert(-2, "witness")
>>> ary
['witness', 'primary', 'normal']
>>> 
```
but unfortunately there is no jinja or ansible filter which allows you to provide a custom function for sorting.  I am investigating

richm: I still can't find a good way to do this.  The simple way will be to use 3 loops - The first loop looks for `primary` and uses `append`, the next loop looks for `witness` and uses `insert(-1)`, and the last loop looks for "other" and uses `insert(0)`.

spetrosi: You're right, I tested this incorrectly. I think the issue is that inserting on -1 index puts the item into next-to-last position in the list. And to put it into the last position, you need to append. This works for me:
```
{% set ag_replicas_sorted = [] %}
{% for item in ansible_play_hosts %}
{%   if hostvars[item]['mssql_ha_replica_type'] == 'primary' %}
{{ ag_replicas_sorted.append(item) -}}
{%   elif hostvars[item]['mssql_ha_replica_type'] == 'witness' %}
{{ ag_replicas_sorted.insert(-1, item) -}}
{%   else %}
{{ ag_replicas_sorted.insert(0, item) -}}
{%   endif %}
{% endfor %}
```
What do you think?

spetrosi: Note that there must be 1 primary and 1 or 0 witness replicas, otherwise SQL Server would fail. This sort should work in any case, howerver.

richm: This will still not work:
```
>>> ary = []
>>> ary.insert(0, "normal")
>>> ary.insert(-1, "witness")
>>> ary.append("primary")
>>> ary
['witness', 'normal', 'primary']
```

spetrosi: I see now, you're so right. Is it fine to do this in two loops to sort primary and witness in a single loop, and then add synchronous in a second loop?
```jinja
{% set ag_replicas_sorted = [] %}
{% for item in ansible_play_hosts %}
{%   if hostvars[item]['mssql_ha_replica_type'] == 'primary' %}
{{ ag_replicas_sorted.append(item) -}}
{%   elif hostvars[item]['mssql_ha_replica_type'] == 'witness' %}
{{ ag_replicas_sorted.insert(0, item) -}}
{%   endif %}
{% endfor %}
{% for item in ansible_play_hosts %}
{%   if 
       (hostvars[item]['mssql_ha_replica_type'] not in ['primary','witness'])
       and
       (hostvars[item]['mssql_ha_replica_type'] not in ag_replicas_sorted)
%}
{{ ag_replicas_sorted.insert(0, item) -}}
{%   endif %}
{% endfor %}
```

richm: Yes, I think that will work.

spetrosi:
I hurried and merged without pushing my last commit. This work is to be continued in #99.

PR #101, opened 2022-08-23
Unset the __mssql_sqlcmd_input_file variable

spetrosi:
This is required because in the case when a task that precedes the
inputtask fails, the print task prints a previous result


spetrosi:
[citest pending]

PR #104, opened 2022-08-24
Remove mssql_ha_firewall_configure, rename mssql_firewall_configure to mssql_manage_firewall

spetrosi:
Previously, two variables existed, mssql_ha_firewall_configure and
mssql_firewall_configure. For simplicity, the role now only uses
mssql_firewall_configure for configuring firewall.

mssql_manage_firewall makes more sense because firewall is a service and
the word manage fits it better.
Other system roles use the rolename_manage_firewall wording too and
mssql must be consistent with them for simplicity

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -8,7 +8,7 @@ mssql_upgrade: false
 mssql_password: null
 mssql_edition: null
 mssql_tcp_port: 1433
-mssql_firewall_configure: false
+mssql_manage_firewall: false
 mssql_ip_address: null
 mssql_pre_input_sql_file: null
 mssql_post_input_sql_file: null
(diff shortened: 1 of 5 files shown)


spetrosi:
[citest pending]

defaults/main.yml
```diff
@@ -28,7 +28,6 @@ mssql_rpm_key: https://packages.microsoft.com/keys/microsoft.asc
 mssql_server_repository: "{{ __mssql_server_repository }}"
 mssql_client_repository: "{{ __mssql_client_repository }}"
 mssql_ha_configure: false
-mssql_ha_firewall_configure: false
```
richm, line 31: don't you need `mssql_manage_firewall: false` here?

spetrosi: Thanks, fixed this.

richm: oops - typo

richm (approved):

spetrosi:
[citest pending]

Issue #115, opened 2022-09-07
Error when using 1.2.4


ajith2234:
Hi 
I have been using this role from ansible galaxy for the past few days. I encountered an error when I updated it to the latest version (1.2.4). 
The error occurs in the "TASK [microsoft.sql.server : Distribute certificate and private key to managed nodes]" task. The error is as shown  below:

![image](https://user-images.githubusercontent.com/90108350/188811648-3dd20dca-58e3-401b-b982-bd0e2a3c551c.png)

After the error occurred to the code which was running fine before, I downgrade the role to 1.2.3 and the error was fixed. I am currently using the 1.2.3 version and it works fine but just wanted to inform you about the error faced in the new version.

Thank you
Ajith

spetrosi:
Hi Ajith,

I cannot replicate this issue with 1.2.4, the role runs smoothly for me and everything works.
The version 1.2.4 apparently does not change tasks related to your issue, so that's weird that the behavior differs for you.
Can you provide more details on what OSes you use, and also share your playbook and inventory. I need it to replicate the issue.
If you would have time please also share full output of ansible, it will help troubleshooting too.

spetrosi:
Diff between 1.2.4 and 1.2.3: https://github.com/linux-system-roles/mssql/compare/1.2.3...1.2.4

spetrosi:
@ajith2234 I noticed that you provide the `mssql_pre_input_sql_file` variable incorrectly. The version 1.2.4 updated mssql_pre_input_sql_file, it now requires a list instead of a string like it was before.

My guess is that your primary node fails due to this issue, and then it is not being configured for HA. This is a bug to a role, it should check if the primary replica is still being worked on by Ansible prior to configuring HA.

Try updating the setting fact task 

```
    - name: Set facts to create a test DB on primary as a pre task
      set_fact:
        mssql_pre_input_sql_file: 
          - create_example_db.j2
        __mssql_test_db_name: ExampleDB
      when: mssql_ha_replica_type == 'primary'
```

Note that the `create_example_db.j2` is a template and uses the `__mssql_test_db_name` variable. It is for the role test purposes. You can rather use something more straighforward e.g. https://github.com/linux-system-roles/mssql/blob/master/tests/files/create_ExampleDB1.sql
Also note that the role in version 1.2.4 supports configuring HA cluster without a database too if you do not provide `mssql_ha_db_names`.

[...]

PR #117, opened 2022-09-16
Role must check if primary is available prior to configuring HA

spetrosi:


spetrosi:
Fixing a bug identified in  #115

richm (approved):

PR #118, opened 2022-09-16
Move all ha-related tasks under a single block

spetrosi:


tasks/main.yml
```diff
@@ -553,188 +568,221 @@
...
+        - name: Save credentials for the {{ mssql_ha_login }} SQL Server login
+          copy:
+            content: |
+              {{ mssql_ha_login }}
+              {{ mssql_ha_login_password }}
+            dest: /var/opt/mssql/secrets/passwd
+            owner: root
+            group: root
+            mode: 0400
+            force: true
```
richm, line 772: should have a `no_log: true` here

spetrosi: It should, I should check the role for security everywhere. Add a test with the new vault CI feature

spetrosi:
[citest]

tasks/main.yml
spetrosi: Diff in this PR is very dirty so I will merge it and then continue in future PRs

spetrosi: Fixing in  #119

PR #119, opened 2022-09-19
Add no_log true to tasks listing credentials on output

spetrosi:
For sqlcmd_input_sql_file add tasks to block to print the output of sqlcmd in case it fails. It is now required because the task itself has no_log true.


spetrosi:
[citest]

spetrosi:
[citest]

PR #120, opened 2022-09-20
Fix Reload service daemon task taking 30 min

spetrosi:


richm:
you are deleting several tests?

spetrosi:
> you are deleting several tests?

tests_idempotency_2017.yml and tests_idempotency_2019.yml have tasks that take 30 minutes, I want to first check if the fix works on these tests, I'll return other tests after.

PR #123, opened 2022-10-04
Remove support for Fedora 36

spetrosi:
mssql-server package does not support Fedora at all but it works on Fedora <36. Once Microsoft adds mssql-server package for RHEL 9 it should work on Fedora 36 too.


spetrosi:
[citest]

richm (approved):

PR #124, opened 2022-10-04
Allow *_input_sql_files vars to take lists and strings

### LINKED ISSUE #122: Failure if mssql_pre_input_sql_file isn't defined as a list
When configuring mssql_pre_input_sql_file as a string:

mssql_pre_input_sql_file: create_example_db.j2

The role fails: 
TASK [microsoft.sql.server : Pre-input SQL scripts to SQL Server] ************************************************************************************************************************************************************************************************************************************************************
fatal: [r8-server2]: FAILED! => {"msg": "Invalid data passed to 'loop', it requires a list, got this instead: create_example_db.j2. Hint: If you passed a list/dict of just one element, try adding wantlist=True to your lookup invocation or use q/query instead of lookup."}

The README.md file states that this variable is a string value, and has an example of it being used as I attempted above.  

We should either update the README.md file to make it clear that this variable must be a list, or update the role to allow declaring the variable like this to work.

spetrosi:
resolves #122

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,9 +162,19 @@ Note that this task is not idempotent, the role always inputs an SQL file if any
 
 You can find an example of an SQL file at `tests/sql_script.sql` at the role directory.
 
+You can set these variables to a list of files, or to a string containing single file.
+For example:
+
+```yaml
+mssql_pre_input_sql_file: script0.sql
+mssql_post_input_sql_file:
+  - script1.sql
+  - script2.sql
+```
+
 Default: `null`
 
-Type: `str`
+Type: `string` or `list`
 
 ### `mssql_debug`
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -313,12 +313,18 @@
       when: __mssql_password_query is failed
       notify: Restart the mssql-server service
 
+# Input files var as list even when provided as sting
 # Keep this task after setting password
 - name: Pre-input SQL scripts to SQL Server
   vars:
-    __mssql_sql_files_to_input: "{{ mssql_pre_input_sql_file }}"
+    __mssql_sql_files_to_input: "{{
+      mssql_pre_input_sql_file
+      if mssql_pre_input_sql_file | type_debug == 'list'
+      else
+      [mssql_pre_input_sql_file]
+    }}"
   include_tasks: input_sql_files.yml
-  when: mssql_pre_input_sql_file | length > 0
+  when: mssql_pre_input_sql_file != []
 
 - name: Set a new edition for MSSQL
   when:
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -460,7 +460,8 @@ For example, you can write a `create_example_db.sql` SQL script that creates a t
 ```yaml
 - name: Set facts to create a test DB on primary as a pre task
   set_fact:
-    mssql_pre_input_sql_file: create_example_db.sql
+    mssql_pre_input_sql_file:
```
richm, line 463:
```suggestion
    mssql_pre_input_sql_files:
```

README.md
```diff
@@ -585,8 +586,10 @@ This example shows how to use the role to set up SQL Server and enable the follo
     mssql_install_fts: true
     mssql_install_powershell: true
     mssql_tune_for_fua_storage: true
-    mssql_pre_input_sql_file: myusers.sql
-    mssql_post_input_sql_file: mydatabases.sql
+    mssql_pre_input_sql_file:
```
richm, line 589:
```suggestion
    mssql_pre_input_sql_files:
```

README.md
```diff
@@ -585,8 +586,10 @@ This example shows how to use the role to set up SQL Server and enable the follo
...
+      - myusers.sql
+    mssql_post_input_sql_file:
```
richm, line 591:
```suggestion
    mssql_post_input_sql_files:
```

tasks/main.yml
```diff
@@ -38,12 +38,42 @@
...
+      debug:
+        msg: >-
+          The mssql_pre_input_sql_file variable is deprecated and will be
+          removed in a future version. Edit your playbook to use the
+          mssql_pre_input_sql_files (written with files in plural) variable
+          instead.
+
+    - name: Link the deprecated mssql_pre_input_sql_file fact
+      set_fact:
+        mssql_pre_input_sql_files: "{{ mssql_pre_input_sql_file }}"
```
richm, line 61: If it is possible that the user has specified a list in `mssql_pre_input_sql_file`, then
```yaml
        mssql_pre_input_sql_files: "{{ (mssql_pre_input_sql_file is string) | ternary([mssql_pre_input_sql_file], mssql_pre_input_sql_file) }}"
```
There isn't really a good test for `list` - both `iterable` and `sequence` return `true` for strings - https://jinja.palletsprojects.com/en/2.9.x/templates/#list-of-builtin-tests - so this assumes if it is not a string, it is a list

spetrosi: Hi Rich, please take a look at the last commit `Allow *_input_sql_files vars to take lists and strings`, do you think if making var to take both lists and strings makes sense?

richm: > Hi Rich, please take a look at the last commit `Allow *_input_sql_files vars to take lists and strings`, do you think if making var to take both lists and strings makes sense?

Looks good

README.md
spetrosi: I decided that renaming mssql_pre_input_sql_file to mssql_pre_input_sql_files is an overkill, I'd rather allow mssql_pre_input_sql_file to get both lists and strings

spetrosi:
[citest]

richm (approved):

PR #125, opened 2022-10-06
Use the certificate role to create the cert and the key

nhosoi:
- Introduce a variable mssql_certificates to set the certificate_requests.
- Add the test case to test_tls_2019.yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,31 @@ Default: `null`
 
 Type: `bool`
 
+#### `mssql_tls_certificates`
+
+A certificate and a key for TLS encryption are created using the `certificate` role
+from the `fedora.linux_system_roles` collection.
+
+The value of `mssql_tls_certificates` is set to the variable `certificate_requests`
+in the `certificate` role.
+Please see the `certificate_requests` section in the `certificate` role documentation
+for details.
+
+With this example, a certificate FILENAME.crt is created in `/etc/pki/tls/certs` and
+a key FILENAME.key is in `/etc/pki/tls/private`.
+```yaml
+mssql_tls_certificates:
+  - name: FILENAME
+    dns: *.example.com
+    ca: self-sign
+```
+When `mssql_tls_certificates` is specified, `mssql_tls_cert` and `mssql_tls_private_key`
+are not needed to set.
+
+Default: `[]`
+
+Type: `list of dictionary`
+
 #### `mssql_tls_cert`
 
 Path to the certificate file to copy to SQL Server.
@@ -522,30 +547,6 @@ The role creates an availability group listener using the following values:
 Default: `null`
 
 Type: `string`
-
-#### `mssql_certificates`
-
-A certificate and a key for TLS encryption are created using the `certificate` role.
-
-The value of `mssql_certificates` is set to the variable `certificate_requests` for
-the `certificate` role.
-Please see the `certificate_requests` section in the `certificate` role documentation
-for details.
-
-With this example, a certificate FILENAME.crt is created in `/etc/pki/tls/certs` and
-a key FILENAME.key is in `/etc/pki/tls/private`.
-```yaml
-mssql_certificates:
-  - name: FILENAME
-    dns: *.example.com
-    ca: self-sign
-```
-When `mssql_certificates` is specified, `mssql_tls_cert` and `mssql_tls_private_key`
-are not needed to set.
-
-Default: `[]`
-
-Type: `list of dictionary`
 
 ## Example Playbooks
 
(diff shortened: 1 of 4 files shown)


nhosoi:
[citest]

nhosoi:
@spetrosi, this is not related to this pr, but when I was reading README.md to make the modification for this pr, I noticed there are `Type: string` and `Type: str`. Maybe, we could replace one with another, e.g., `string` with `str`? Or are they supposed to be different?
```
$ egrep "Type:" README.md | egrep string | wc
      9      18     135
$ egrep "Type:" README.md | egrep "\<str\>" | wc
     11      22     132
```

README.md
```diff
@@ -523,6 +523,30 @@ Default: `null`
 
 Type: `string`
 
+#### `mssql_certificates`
+
+A certificate and a key for TLS encryption are created using the `certificate` role.
```
spetrosi, line 528:
```suggestion
Set this variable to generate a certificate and a key for TLS encryption.
The role uses the `fedora.linux_system_roles.certificate` role to generate TLS certificate and key.
```
Replace passive voice with active.
mssql and linux_system_roles are in separate collections and RPMs, so mssql users might not know what we relate to when saying "certificate role". I am using the FQCN to point to roles to make it clearer. For the downstream, I then replace accordingly in the spec file.

README.md
```diff
@@ -523,6 +523,30 @@ Default: `null`
...
+
+The value of `mssql_certificates` is set to the variable `certificate_requests` for
```
spetrosi, line 530: Historically, the mssql role has variables related to tls encryption prefixed with `mssql_tls_`. I think it would be more intuitive to replace `mssql_certificates` with `mssql_tls_certificates`. And place the description of `mssql_tls_certificates` under `mssql_tls_enable` in readme.

README.md
spetrosi: I think I can edit the README to fix style, e.g. use active voice and make it consistent with the rest, after you finish your work, it will save you some time.

README.md
```diff
@@ -1035,6 +1059,20 @@ This example playbooks sets the `firewall` variables for the `fedora.linux_syste
     - microsoft.sql.server
 ```
 
+### Setting Up SQL Server with TLS Encryption using self signed certificate and key created by the certificate role
+
+    - name: Test with certs created by the certificate role
+      include_role:
+        name: linux-system-roles.mssql
+        public: true
```
spetrosi, line 1067: Is setting `public: true` strictly required?

tasks/main.yml
```diff
@@ -477,6 +477,26 @@
...
+          vars:
+            certificate_requests: |
+              {% for _cert in mssql_certificates %}
+              {%   set _ = _cert.__setitem__('owner', 'mssql') %}
+              {%   set _ = _cert.__setitem__('group', 'mssql') %}
+              {%   set _ = _cert.__setitem__('mode', '0600') %}
+              {% endfor %}
+              {{ mssql_certificates }}
+
+        - name: Generate certificates using the certificate role
```
spetrosi, line 493: Wrong task name

tasks/main.yml
```diff
@@ -477,6 +477,26 @@
...
+          set_fact:
+            mssql_tls_cert: "{{ mssql_certificates.0.name }}.crt"
+            mssql_tls_private_key: "{{ mssql_certificates.0.name }}.key"
+
+      when: mssql_certificates | length > 0
```
spetrosi, line 498: Someone said that the latest version of ansible-lint was complaining about putting `when` in blocks underneath tasks. I think it is more readable when `when` is set after the `block` statement.

tasks/main.yml
```diff
@@ -477,6 +477,26 @@
   when:
     - mssql_tls_enable is not none
   block:
+    - block:
```
spetrosi, line 480: I was putting names for blocks everywhere, I thought it is a requirement, I think it's fine without a name too though as tasks will be explaining what's going on anyway. I am afraid that newer versions of ansible-lint might complain about this too.

README.md
richm: probably not - I'm assuming this is a copy/paste from a test

nhosoi:
[citest]

spetrosi:
Thanks @nhosoi, all looks good to me, I applied by changes to README, I also modified tests to remove redundant flush handlers. Please review my commit and I think you can merge.

spetrosi:
[citest]

README.md
```diff
@@ -259,46 +253,42 @@ Type: `bool`
...
-
-Type: `str`
+Paths to the certificate and private key files to copy to SQL Server.
 
-#### `mssql_tls_private_key`
+You are responsible for creating and securing TLS certificate and private key files.
+It is assumed you have a CA that can issue these files.
 
-Path to the private key file to copy to SQL Server.
+When you use this variable, the role copies TLS cert and private key files to SQL Server and configures SQL Server to use these files to encrypt connections.
```
nhosoi, line 282: ^this variable^these variables^ ? ;)

spetrosi:
[citest]

PR #141, opened 2022-11-28
weekly-ci: do not create a new PR every time

ukulekek:


.github/workflows/weekly_ci.yml
```diff
@@ -8,32 +8,11 @@ env:
   BRANCH_NAME: "weekly-ci"
   COMMIT_MESSAGE: "Weekly CI check"
   COMMENT: "[citest]"
+  MAIN_BRANCH: "main"
```
richm, line 11: please don't hardcode the main branch name - we will be changing the remaining `master` branches to `main` at some point in the near future - instead, please use this method for determining the main branch name - https://github.com/linux-system-roles/auto-maintenance/blob/master/role-make-version-changelog.sh#L42

ukulekek: Thank you, I added the function to the script

.github/workflows/weekly_ci.yml
```diff
@@ -12,56 +12,66 @@ jobs:
...
+          git rebase $(get_main_branch)
           git commit --allow-empty -m "${{ env.COMMIT_MESSAGE }}"
-          git push --set-upstream origin ${{ env.BRANCH_NAME }}
+          git push -f --set-upstream origin ${{ env.BRANCH_NAME }}
 
       - name: Create and comment pull request
         uses: actions/github-script@v6
         with:
-          github-token: ${{ secrets.GH_PUSH_TOKEN }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
richm, line 49: I think the comment still has to be created by the user which has the token `GH_PUSH_TOKEN` - otherwise, the citest comment won't trigger ci

ukulekek: Oh, sorry, this is leftover, will change it back. Thanks for noticing

richm (approved):

PR #153, opened 2022-12-14
Set password to sqlcmd with env variable for security

spetrosi:


spetrosi:
[citest]

tasks/main.yml
```diff
@@ -285,11 +285,11 @@
   block:
     - name: Prepare MSSQL and facts for logging in
       include_tasks: verify_password.yml
-      vars:
-        __mssql_password: "{{ mssql_password }}"
 
     - name: Check if the set password matches the existing password
       command: "{{ __mssql_sqlcmd_login_cmd }} -Q 'SELECT @@VERSION'"
+      environment:
+        - SQLCMDPASSWORD: "{{ mssql_password | quote }}"
```
richm, line 292: does it need to be quoted?  not sure how environment variables work in ansible, but if you do something like in python `os.environ["SQLCMDPASSWORD"] = somevalue` then `somevalue` does not have to be shell quoted

PR #161, opened 2023-01-12
Restructure README files to split it into available scenarios

spetrosi:
With more configuration scenarios added, README file became messy. These changes split README file into sections, each describing a specific user story and providing example playbooks in-place.


spetrosi:
The sections are easy to navigate with a table of content that can be created for HTML version of the README. I will submit a PR for that soon.

PR #162, opened 2023-01-12
ansible-lint 6.x fixes

richm:
The big one is that ansible-lint doesn't like templates in `name`
strings except at the end.  In general, Ansible does not like having
templated variables in `name` values because it makes it harder to
grep the source to find a log message in the source.
The other ones are jinja spacing cleanup, use of `true`/`false`
instead of `yes`/`no`, and various other cleanup.


richm:
I also had to move `tests/clean_up_mssql_pacemaker.yml` to `tests/playbooks/clean_up_mssql_pacemaker.yml` - our default schema for ansible-lint assumes files in `tests/` not matching `tests_*.yml` are tasks files, not playbooks.

richm:
[citest]

spetrosi (approved):
lgtm

PR #165, opened 2023-01-25
Create separate github actions for various checks; get rid of monolithic tox.yml

richm:
There are now separate github actions for the various tests, instead of all of
them being done as part of the tox tests in tox.yml - ansible-lint,
ansible-test, etc.

Use the officially supported github actions e.g. for ansible-lint, ansible-test,
rather than using our own from `tox-lsr` and trying to keep up with the latest
changes.  Developers will still be able to use `tox-lsr` on their local
development environments to run these tests in the same way that they are run in
github actions, so that errors found when submitting PRs can be reproduced and
corrected locally without too many github UI roundtrips.

Using separate github actions, and especially the official github actions which
generally have support for in-line comments, should help greatly with
readability and troubleshooting test results.

Python tests are removed from roles that do not use python.

Python tests are now done by python-unit-tests.yml which also does the black,
flake8, and pylint tests.


spetrosi:
ansible-lint gets confused about the `.collection` directory that is used when building collection, and searches changelog in it, and fails. I am suppressing the changelog rule because CHANGELOG.Md exists in the root directory.
Error message:
```
galaxy[no-changelog]: No changelog found. Please add a changelog file.
Refer to the galaxy.md file for more info.
.collection/galaxy.yml:1
```

spetrosi:
@richm by the way the error I mentioned above cannot be reproduced with `tox -e ansible-lint`. GitHub action probably use newer ansible-lint version.

richm:
> @richm by the way the error I mentioned above cannot be reproduced with `tox -e ansible-lint`. GitHub action probably use newer ansible-lint version.

It should work if you use tox-lsr from main branch - https://github.com/linux-system-roles/tox-lsr/blob/main/src/tox_lsr/config_files/tox-default.ini#L254
This is the same as the action https://github.com/ansible/ansible-lint-action/blob/main/action.yml#L26

spetrosi:
Do you approve my commit @richm ?

spetrosi:
I run update to latest tox-lsr using `pip install --user git+https://github.com/linux-system-roles/tox-lsr@main -U`, after this running `tox -e ansible-lint` on your commit in this branch succeeds, while in the CI it fails. I'll try to investigate further tomorrow.

.github/workflows/changelog_to_tag.yml
```diff
@@ -74,3 +75,8 @@ jobs:
           body_path: ./.tagmsg.txt
           draft: false
           prerelease: false
+      - name: Publish role to Galaxy
```
richm, line 78: The mssql role doesn't currently have this task to publish the role to Galaxy - is that because we don't publish a legacy role, but only publish a collection?

spetrosi: Yes exactly

spetrosi (commented):
lgtm

PR #166, opened 2023-01-25
Rename mssql_ha_listener_port to mssql_ha_log_transport_port

spetrosi:
As per feedback from Microsoft, mssql_ha_listener_port should be called HA communication port (or HA_log_transport_port),  as that port is used for the transaction log replication between primary and secondary replica. Listener is a term used for AG listener associated with AG which is used to route client connection to primary replica (or read only secondary replica based on configuration and request type).

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,7 +535,7 @@ Setting to `false` does not remove configuration for high availability.
 When set to `true`, the role performs the following tasks:
 
 1. Include the `fedora.linux_system_roles.firewall` role to configure firewall:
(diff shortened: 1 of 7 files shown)


spetrosi:
Microsoft folks said "It should ideally be called HA communication port (or HA_log_transport_port)", so I will clarify further

richm:
other than the line-length issue, lgtm

spetrosi:
[citest]

PR #192, opened 2023-04-13
Add Commitlint GitHub action

spetrosi:


spetrosi:
Adding this in https://github.com/linux-system-roles/.github/pull/24 it will be applied globally to all roles with automation.

PR #199, opened 2023-05-10
feat: Support custom data and logs storage paths

spetrosi:
Previously, the role was configuring the default data and logs storage
paths. Currently, you can provide custom storage paths with variables
`mssql_datadir` and `mssql_logdir`, and optionally set permissions for
the custom paths with `mssql_datadir_mode` and `mssql_logdir_mode`
variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ The role also optimizes the operating system to improve performance and throughp
 
 ## Role Scenarios
 
+This section describes different scenarios that the role can accomplish.
+You can combine the scenarios by combining variables serving different scenarios into a larger playbook.
+
 ### Configuring General SQL Server Settings
 
 These variables apply to general SQL Server configuration.
@@ -293,6 +296,60 @@ Type: `string`
 The URL to the Microsoft production repository.
 
 Default: `{{ __mssql_client_repository }}`
+
+Type: `string`
+
+### Configuring Custom Data and Logs Storage Paths
+
+Optional: Use these variables to configure SQL Server to store data and logs in custom paths.
+
+#### Variables
+
+##### mssql_datadir
+
+The path to the directory that SQL Server must use to store data.
+When defined, the role creates the provided directory and ensures correct permissions and ownership for it.
+
+Note that if you change this path, the previously used directory and all it's content remains at the original path.
+
+Default: `null`
+
+Type: `string`
+
+##### mssql_datadir_mode
+
+The permissions to be set for the `mssql_datadir` path in the format of the Ansible `file` module `mode` variable.
+
+Quote the mode like `'0700'` so Ansible parses it as a string to avoid conflicts with octal numbers.
+
+If mode is not specified and the destination directory *does not* exist, the role uses the default umask on the system when setting the mode.
+If mode is not specified and the destination directory *does* exist, the role uses the mode of the existing directory.
+
+Default: `null`
+
+Type: `string`
+
+##### mssql_logdir
+
+The path to the directory that SQL Server must use to store logs.
+When defined, the role creates the provided directory and ensures correct permissions and ownership for it.
+
+Note that if you change this path, the previously used directory and all it's content remains at the original path.
+
+Default: `null`
+
+Type: `string`
+
+##### mssql_logdir_mode
+
+The permissions to be set for the `mssql_logdir` path in the format of the Ansible `file` module `mode` variable.
+
+Quote the mode like `'0700'` so Ansible parses it as a string to avoid conflicts with octal numbers.
+
+If mode is not specified and the destination directory *does not* exist, the role uses the default umask on the system when setting the mode.
+If mode is not specified and the destination directory *does* exist, the role uses the mode of the existing directory.
+
+Default: `null`
 
 Type: `string`
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -10,16 +10,19 @@ mssql_edition: null
 mssql_tcp_port: 1433
 mssql_manage_firewall: false
 mssql_ip_address: null
-
-mssql_pre_input_sql_file: []
-mssql_post_input_sql_file: []
-mssql_debug: false
-
 mssql_enable_sql_agent: null
 mssql_install_fts: null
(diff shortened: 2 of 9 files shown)


spetrosi:
[citest]

tasks/main.yml
```diff
@@ -256,6 +256,27 @@
...
+- name: Ensure the directory and permissions {{ item }}
+  when: item is not none
+  loop:
+    - "{{ mssql_datadir }}"
+    - "{{ mssql_logdir }}"
+  file:
+    path: "{{ item }}"
+    state: directory
+    owner: mssql
+    group: mssql
```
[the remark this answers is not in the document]
spetrosi: The mode of the dirs created by default - `/var/opt/mssql/data` and `/var/opt/mssql/logs` - is `0700`. Mode is the only additional parameter here that users might want to configure, owner and group must be `mssql`. So probably it would be easier to add additional variables `mssql_datadir_mode` and `mssql_logdir_mode` that default to 0700 and can be customized.

spetrosi:
[citest]

richm (approved):

spetrosi:
[citest]

spetrosi:
[citest]

README.md
```diff
@@ -296,6 +299,57 @@ Default: `{{ __mssql_client_repository }}`
...
+Type: `string`
+
+##### mssql_logdir_mode
+
+The permissions to be set for the `mssql_logdir` path in the format of the Ansible `file` module `mode` variable.
+
+Quote the mode like `'0700'` so Ansible parses it as a string to avoid conflicts with octal numbers.
+
+If mode is not specified and the destination directory *does not* exist, the role uses the default umask on the system when setting the mode.
+If mod' is not specified and the destination directory *does* exist, the role uses the mode of the existing directory.
```
richm, line 347:
```suggestion
If mode is not specified and the destination directory *does* exist, the role uses the mode of the existing directory.
```

defaults/main.yml
```diff
@@ -10,16 +10,19 @@ mssql_edition: null
...
-mssql_post_input_sql_file: []
-mssql_debug: false
-
 mssql_enable_sql_agent: null
 mssql_install_fts: null
 mssql_install_powershell: null
 mssql_enable_ha: null
 mssql_tune_for_fua_storage: null
+mssql_datadir: null
+mssql_datadir_mode: null
```
richm, line 19: I think you need to have a default value here, either `"0700"` or `"0750"` or  `"0755"` - whatever mssql recommends

spetrosi:
[citest]

spetrosi:
[citest_bad]

defaults/main.yml
spetrosi: In this case, the role would overwrite the existing mode back to default if the correct mode is not set with `mssql_datadir_mode`, may be this is expected from Ansible, but I do not think that this is a good UX. I like to be it like this:
If users provide `mssql_datadir_mode`, it is used,
If users don't provide  `mssql_datadir_mode` and the `mssql_datadir` directory exists, the existing mode is kept,
If users don't provide  `mssql_datadir_mode` and the `mssql_datadir` directory doesn't exist, the default mode is used.

There is no recommendation from Microsoft on this, by default the directory is created with `0755`, which is just the default umask on the machine.

richm: ok

tasks/configure_storage_paths.yml
```diff
@@ -0,0 +1,31 @@
...
+
+- name: Ensure the directory and permissions for {{ __mssql_storage_path }}
+  # null converts to an empty string when setting vars in the parent task
+  when: __mssql_storage_mode != ''
+  file:
+    path: "{{ __mssql_storage_path }}"
+    state: directory
+    owner: mssql
+    group: mssql
+    mode: "{{ __mssql_storage_mode }}"
```
richm, line 16: I think you can do this with a single `file` module
```yaml
    mode: "{{ __mssql_storage_mode is None | ternary(
      __mssql_storage_path_stat is exists | ternary(
       __mssql_storage_path_stat.stat.mode, '0700'),
      __mssql_storage_mode) }}"
```
or if you prefer the more verbose version
```yaml
    mode: "{{ __mssql_storage_mode
      if __mssql_storage_mode is not None else
      __mssql_storage_path_stat.stat.mode if
      __mssql_storage_path_stat is exists else
      '0700' }}"
```

richm: or if you want the default mode to be configurable, then add that as a parameter to configure_storage_paths.yml e.g.
```yaml
    mode: "{{ __mssql_storage_mode
      if __mssql_storage_mode is not None else
      __mssql_storage_path_stat.stat.mode if
      __mssql_storage_path_stat is exists else
      __mssql_storage_mode_default }}"
```

spetrosi: Oh, that's good, I didn't know that you can do `mode: ""` and the file module will consider this an unset mode, hence I created another task with mode literally not set.

`__mssql_storage_path_stat is exists` doesn't work for me, see below. I used `__mssql_storage_path_stat.stat.exists` instead.

```
TASK [linux-system-roles.mssql : debug] *************************************************************************************************************************************
ok: [127.0.0.3] => {
    "__mssql_storage_path": "/tmp/log"
}

TASK [linux-system-roles.mssql : debug] *************************************************************************************************************************************
ok: [127.0.0.3] => {
    "__mssql_storage_path_stat.stat.path": "/tmp/log"
}

TASK [linux-system-roles.mssql : debug] *************************************************************************************************************************************
ok: [127.0.0.3] => {
    "__mssql_storage_path_stat.stat.path is exists": false
}

TASK [linux-system-roles.mssql : debug] *************************************************************************************************************************************
ok: [127.0.0.3] => {
    "__mssql_storage_path_stat.stat.exists": true
}

TASK [linux-system-roles.mssql : debug] *************************************************************************************************************************************
ok: [127.0.0.3] => {
    "__mssql_storage_path is exists": false

```

richm (approved):

spetrosi:
[citest]

PR #201, opened 2023-05-25
refactor: Refactor tests and fix tests timeouts

spetrosi:


spetrosi:
[citest]

tasks/sqlcmd_input_file.yml
```diff
@@ -42,11 +42,15 @@
         or
         (__mssql_sqlcmd_input is failed)
       block:
-        - name: >-
-            Print the output of the sqlcmd command
-            if not empty for {{ item }}
+        - name: Print the output of the sqlcmd command for {{ __mssql_debug }}
+          when: "{{ __mssql_debug }} | length > 0"  # noqa no-jinja-when
```
richm, line 46:
```suggestion
          when: __mssql_debug | length > 0
```
unless I'm missing something?

tasks/sqlcmd_input_file.yml
```diff
@@ -42,11 +42,15 @@
...
           debug:
-            var: __mssql_sqlcmd_input.stdout_lines
+            var: "{{ __mssql_debug }}"
```
richm, line 48: ?

tasks/sqlcmd_input_file.yml
richm: does `__mssql_debug` hold the name of another variable you want to expand?  If so, I think using the `vars` lookup https://docs.ansible.com/ansible/2.9/plugins/lookup/vars.html is a better way to do this.

tasks/sqlcmd_input_file.yml
```diff
@@ -43,9 +43,9 @@
         (__mssql_sqlcmd_input is failed)
       block:
         - name: Print the output of the sqlcmd command for {{ __mssql_debug }}
-          when: "{{ __mssql_debug }} | length > 0"  # noqa no-jinja-when
+          when: "{{ lookup('vars', '__mssql_debug') }} | length > 0"  # noqa no-jinja-when
```
richm, line 46: what is the value of `__mssql_debug`?  is it the output of sqlcmd?

tasks/sqlcmd_input_file.yml
```diff
@@ -42,11 +42,15 @@
         or
         (__mssql_sqlcmd_input is failed)
       block:
-        - name: >-
-            Print the output of the sqlcmd command
-            if not empty for {{ item }}
+        - name: Print the output from inputting {{ __mssql_sqlcmd_loop }}
```
richm, line 45: why not just
```yaml
        - name: Print the output from inputting {{ item }}
          debug:
            msg: |
              STDOUT
              {{ __mssql_sqlcmd_input.stdout }}
              STDERR
              {{ __mssql_sqlcmd_input.stderr }}
```
?

spetrosi:
[citest]

tasks/sqlcmd_input_file.yml
spetrosi: It doesn't print the output as pretty as `debug: var:`, I now simplified the whole thing and removed a redundant block. The loop is still there though, with it the output and the filename are ### printed separately, which is easier to read.

richm (approved):

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
Apparently, raising delay from 3 to 15 in https://github.com/linux-system-roles/mssql/pull/201/files#diff-5770d5488125004425452c7c8412bc3ee74452473cf84224fd42a0aa9220789cR97 helped with periodical issues with mssql-server not available when verifying password in tests. Raising to 10 didn't help. So far, this wait_for task with delay 15 runs 20 times in all tests, which sums up to 3 minutes, which is fine.
The issue happened around 3 times of all test runs on all OSes, so it is hard to troubleshoot this. And only happens in the test task `tests/tasks/verify_settings.yml`, not in the role itself. Most probably some issue with mssql-server.

PR #207, opened 2023-06-14
ci: Add cleanup

spetrosi:
Enhancement: Add cleanup task to all tests 

Reason: Our CI is now capable of running tests in series on a single node to make tests faster. This requires a cleanup 

Result: Each test runs a complete cleanup of mssql and related services at it's end


spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

tests/tests_2019_upgrade.yml
```diff
@@ -7,6 +7,9 @@
     mssql_accept_microsoft_cli_utilities_for_sql_server_eula: true
     mssql_accept_microsoft_sql_server_standard_eula: true
   tasks:
+    - name: Clean up after the role invocation
```
richm, line 10: The role hasn't been invoked yet?

spetrosi: I found that in our scenario of serial tests invocations in the case when tests fail they don't clean up after themselves (unless you put cleanup task as a rescue in a block, which is ugly with indentation and all. So cleaning before the test invocation would be easier. How does this sound to you?
And the task name yeah should be updated, I'll fix this.

richm: > I found that in our scenario of serial tests invocations in the case when tests fail they don't clean up after themselves (unless you put cleanup task as a rescue in a block, which is ugly with indentation and all.

Understood, but that's how most of the other roles that implement cleanup work - they run everything in a `block`, then use the `always` to run the cleanup.

> So cleaning before the test invocation would be easier. How does this sound to you? And the task name yeah should be updated, I'll fix this.

The other purpose of cleanup is to ensure that the test leaves the system in the same state as when it started, so that tests for other roles can start with a clean system.

re: ugly indentation - yes - tools like vscode with block indentation can help

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

PR #211, opened 2023-07-11
feat: Add mssql_ad_join and mssql_ad_kerberos_user to allow using other domain join methods

### LINKED ISSUE #210: Allow other methods to domain join system
https://github.com/linux-system-roles/mssql/blob/d49168ddac9998a11dc8a47e0fae4f041dc51ef8/tasks/main.yml#L661-L663

It would be preferable to add a variable to optionally disable using the ad_integration role. There are other methods to domain join a system that would be broken if that role were to run. A warning can be added to state that domain joining must be complete before running this role if the variable is set to false.

spetrosi:
Enhancement:
1. Add the `mssql_ad_join` variable to allow using other domain join methods.
2. Add the `mssql_ad_kerberos_user` and `mssql_ad_kerberos_password` variables to allow users to specify a user to obtain kerberos ticket for in the case that default user doesn't match.

Reason: Sometimes, users use other methods to manage joining to AD that would be broken if that role were to run.

Result:
1. Users can set `mssql_ad_join` to `false` and join managed nodes to AD themselves prior to running the role. By default, `mssql_ad_join` = `true`, so the current behavior is not changed.
2. Users can set `mssql_ad_kerberos_user` and `mssql_ad_kerberos_password` to obtain a kerberos ticket for a specific use if the user that the role selects by default doesn't work for them.

Resolves #210

### DIFF
--- a/README.md
+++ b/README.md
@@ -1208,8 +1208,8 @@ This role uses the `fedora.linux_system_roles.ad_integration` role to join SQL S
 
 To join to AD server, you must also provide the following variables:
 * `ad_integration_realm`
-* `ad_integration_password`
 * `ad_integration_user`
+* `ad_integration_password`
 * Optional, You can configure DNS using ad_integration role by providing the following variables:
   ```
   ad_integration_manage_dns: true
@@ -1219,6 +1219,8 @@ To join to AD server, you must also provide the following variables:
   ```
 * Optional: You can provide further variables for the `fedora.linux_system_roles.ad_integration` role if you need.
 
+Optional: If you have already joined managed nodes to AD and you want to skip running the ad_integration role, you can set the [`mssql_ad_join`](#mssql_ad_join) variable to false and omit the `ad_integration_` variables.
+
 #### Prerequisites
 
 Ensure that your AD Server and Linux host meet the prerequisites for joining.
@@ -1226,35 +1228,41 @@ For more information, see [Join SQL Server on a Linux host to an Active Director
 
 #### Finishing AD Server Configuration
 
-After you execute the role to configure AD Server authentication, you must complete one of the following procedures to add AES128 and AES256 kerberos encryption types to the [`mssql_ad_sql_user_name`](#mssql_ad_sql_user_name) on AD Server.
+1. After you execute the role to configure AD Server authentication, you must complete one of the following procedures to add AES128 and AES256 kerberos encryption types to the [`mssql_ad_sql_user_name`](#mssql_ad_sql_user_name) on AD Server.
 
-* For the web UI users, complete the following steps:
-    1. Log in to your AD Server
-    2. Navigate to **Tools** > **Active Directory Users and Computers** > ***domain.com*** > **Users** > ***sqluser*** > **Account**
-    3. In the **Account options** list, select **This account supports Kerberos AES 128 bit encryption** and **This account supports Kerberos AES 256 bit encryption**
-    4. Click **Apply**
+  * For the RSAT UI users, complete the following steps:
+      1. Launch **Active Directory Users and Computers**
+      2. Navigate to the SQL User (Default: ***domain*** > **Users** > ***sqluser***)
+      3. Right click the SQL User account and select **Properties**
+      4. In the **Account** tab, select **This account supports Kerberos AES 128 bit encryption** and **This account supports Kerberos AES 256 bit encryption**
+      5. Click **Apply**
 
-* For the PowerShell users, enter the following command:
-    ```powershell
-    Set-ADUser -Identity <sqluser> -KerberosEncryptionType AES128,AES256
-    ```
+  * For the PowerShell users, enter the following command:
+
+      ```powershell
+      Set-ADUser -Identity <sqluser> -KerberosEncryptionType AES128,AES256
+      ```
 
 #### Verifying Authentication
 
 After you execute the role to configure AD Server authentication and complete [Post Configuration Tasks](#post-configuration-tasks), you log in using Azure Data Studio or complete the following procedure to verify that you can log in to SQL Server from your Linux machine using the <sqluser> account.
 
 1. SSH into the _<sqluser>@<domain.com>_ user on your Linux _client.domain.com_ machine:
-  ```
-  ssh -l <sqluser>@<domain.com> <client.domain.com>
-  ```
+
+    ```
+    ssh -l <sqluser>@<domain.com> <client.domain.com>
+    ```
+
 2. Obtain Kerberos ticket for the Administrator user:
-  ```
-  kinit Administrator@<DOMAIN.COM>
-  ```
+    ```
+    kinit Administrator@<DOMAIN.COM>
+    ```
+
 3. Use `sqlcmd` to log in to SQL Server and, for example, run the query to get current user:
-  ```
-  /opt/mssql-tools/bin/sqlcmd -S. -Q 'SELECT SYSTEM_USER'
-  ```
+
+    ```
+    /opt/mssql-tools/bin/sqlcmd -S. -Q 'SELECT SYSTEM_USER'
+    ```
 
 #### Variables
 
@@ -1263,7 +1271,31 @@ After you execute the role to configure AD Server authentication and complete [P
 Set this variable to `true` to configure for AD Server authentication.
 Setting to `false` does not remove configuration for AD Server authentication.
 
+For an example playbook, see [Configuring AD integration with a pre-created keytab file](#configuring-ad-integration-with-general-parameters).
+
 Default: `false`
+
+Type: `bool`
+
+##### mssql_ad_join
+
+Optional: If you have already joined managed nodes to AD and you want to skip running the ad_integration role, you can set this variable to `false`.
+
+For an example playbook, see [Configuring AD integration with a pre-created keytab file](#configuring-ad-integration-without-joining-to-ad).
+
+Default: `true`
+
+Type: `bool`
+
+##### mssql_ad_obtain_kerberos
+
+Optional: If you have already obtained Kerberos ticket on managed nodes and you want to skip obtaining it, you can set this variable to `false`.
+
+By default, when mssql_ad_obtain_kerberos is set to `true`, you must set `ad_integration_user` and `ad_integration_password` variables.
+
+For an example playbook, see [Configuring AD integration with a pre-created keytab file](#configuring-ad-integration-without-joining-to-ad-and-without-obtaining-kerberos-ticket).
+
+Default: `true`
 
 Type: `bool`
 
@@ -1272,6 +1304,8 @@ Type: `bool`
 Optional: You can use this variable if you don't want the role to generate a keytab file and instead want to pass a pre-created keytab file to mssql-server.
 
 A keytab file must be pre-created on AD Server as per [Tutorial: Use Active Directory authentication with SQL Server on Linux](https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-active-directory-authentication?view=sql-server-ver16). You must ensure that the keytab file is created for the managed node's hostname and for the user provided with the `mssql_ad_sql_user_name` variable.
+
+Note that kerberos ticket is not required when the role uses keytab, therefore when you set `mssql_ad_keytab_file: false` you can also set `mssql_ad_obtain_kerberos: false`.
 
 For an example playbook, see [Configuring AD integration with a pre-created keytab file](#configuring-ad-integration-with-a-pre-created-keytab-file).
 
@@ -1352,7 +1386,7 @@ Type: `string`
 
 #### Example Playbooks
 
-##### Configuring AD integration and generating a keytab file
+##### Configuring AD integration with general parameters
 
 ```yaml
 - name: Configure with AD server authentication
@@ -1369,8 +1403,8 @@ Type: `string`
     mssql_ad_sql_user_name: sqluser
     mssql_ad_sql_password: "p@55w0rD1"
     ad_integration_realm: domain.com
-    ad_integration_password: Secret123
     ad_integration_user: Administrator
+    ad_integration_password: Secret123
     ad_integration_manage_dns: true
     ad_integration_dns_server: 1.1.1.1
     ad_integration_dns_connection_name: eth0
@@ -1379,8 +1413,10 @@ Type: `string`
 
 ##### Configuring AD integration with a pre-created keytab file
 
+If you received a pre-created keytab file and want the role to use it, set variables like in this example:
+
 ```yaml
-- name: Configure with AD server authentication
+- name: Configure with AD server authentication with a pre-created keytab file
   hosts: all
   vars:
     mssql_accept_microsoft_odbc_driver_17_for_sql_server_eula: true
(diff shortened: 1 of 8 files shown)


spetrosi:
@JasonN3 Thank you for submitting #210, please review this, mainly changes to README.md.

spetrosi:
[citest]

richm (approved):

JasonN3:
@spetrosi `ad_integration_user` and `ad_integration_password` are really only needed if the user specified in `mssql_ad_sql_user_name` does not exist. Most admins I've worked with will want to create that ahead of time themselves along with setting the SPN. If you use `mssql_ad_sql_user_name` and `mssql_ad_sql_password` as default values for those variables, it will still allow you to check that the user exists with the same commands, which could also just be skipped in that case.

If you were to allow `ad_integration_user` and `ad_integration_password` to be omitted by using `mssql_ad_sql_username` and `mssql_ad_sql_passwd` as the defaults, you could skip lines 702-742 in tasks/main.yml with the condition that `ad_integration_user` was omitted. The username and password would already be verified when kinit was run on line 689. The rest should be able to succeed just fine. There might be an error when adding the SPN, but that would just be an error to tell them they skipped a step they wanted to do externally to this role.

spetrosi:
Another scenario is to pass only a pre-created keytab file to mssql-server as described in https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-active-directory-authentication?view=sql-server-ver16 steps 3+. I think this is the one that more experienced users would be interested in. This is not.
And for the scenario in this PR, I would like to consciously provide ad_integration_user and ad_integration_password so that the role ensures that required properties and fails if not. It's not a big burden on the user, and it makes the process less error prone. WDYT?

spetrosi:
ad_integration_user is also required to create this user in SQL Server in https://github.com/linux-system-roles/mssql/blob/main/tasks/main.yml#L848
So it's good if the role ensures that is it is present.

spetrosi:
> > Another scenario is to pass only a pre-created keytab file to mssql-server as described in https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-active-directory-authentication?view=sql-server-ver16 steps 3+. I think this is the one that more experienced users would be interested in. This is not. And for the scenario in this PR, I would like to consciously provide ad_integration_user and ad_integration_password so that the role ensures that required properties and fails if not. It's not a big burden on the user, and it makes the process less error prone. WDYT?
> 
> Being able to optionally specify the keytab file would make it more flexible. None of the organizations I've worked with allowed the DBAs to have access to an account that could modify any AD users. If a DBA is running this Ansible role, they would need to be able to fill something in to there and the ad sql user would be enough permissions to verify that the AD admins completed the work correctly. It would only have an error if something wasn't pre-staged correctly and it would give enough information in the error that the DBA could tell the AD admin they didn't pre-stage the account correctly. Least privileged access and separation of duty are core security practices that should be accounted for if you want people to use this role.

So is it fine to require users to provide ad_integration_realm, ad_integration_user, and ad_integration_realm when mssql_ad_join is false? 
I think we could make it so that if user provides mssql_ad_join: false and `mssql_ad_keytab_file`(I will implement it soon) then the only required variable would be `ad_integration_user`. See below why ad_integration_user is needed.

> > ad_integration_user is also required to create this user in SQL Server in https://github.com/linux-system-roles/mssql/blob/main/tasks/main.yml#L848 So it's good if the role ensures that is it is present.
> 
> Why are you creating the user account for `ad_integration_user`? If that account has the ability to add computers to the domain and create and modify user accounts, it should not be an account that you're logging in to the SQL database with. Again, least privileged access needs to be a core concept that is encouraged. You can give options to ignore LPA, but it shouldn't be the default or the only option.

Per the documentation in https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-active-directory-authentication?view=sql-server-ver16#createsqllogins or https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-ad-auth-adutil-tutorial?view=sql-server-ver16#create-active-directory-based-sql-server-logins-in-transact-sql
I specifically verified with Microsoft developer that here the login with the username of ad_integration privileged user must be created, so `ad_integration_user`.

JasonN3:
> > > Another scenario is to pass only a pre-created keytab file to mssql-server as described in https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-active-directory-authentication?view=sql-server-ver16 steps 3+. I think this is the one that more experienced users would be interested in. This is not. And for the scenario in this PR, I would like to consciously provide ad_integration_user and ad_integration_password so that the role ensures that required properties and fails if not. It's not a big burden on the user, and it makes the process less error prone. WDYT?
> > 
> > 
> > Being able to optionally specify the keytab file would make it more flexible. None of the organizations I've worked with allowed the DBAs to have access to an account that could modify any AD users. If a DBA is running this Ansible role, they would need to be able to fill something in to there and the ad sql user would be enough permissions to verify that the AD admins completed the work correctly. It would only have an error if something wasn't pre-staged correctly and it would give enough information in the error that the DBA could tell the AD admin they didn't pre-stage the account correctly. Least privileged access and separation of duty are core security practices that should be accounted for if you want people to use this role.
> 
> So is it fine to require users to provide ad_integration_realm, ad_integration_user, and ad_integration_realm when mssql_ad_join is false? I think we could make it so that if user provides mssql_ad_join: false and `mssql_ad_keytab_file`(I will implement it soon) then the only required variable would be `ad_integration_user`. See below why ad_integration_user is needed.

When `mssql_ad_join` is false, `ad_integration_realm` is going to be needed when verifying all of the pre-staging is done right, so, yes, that should still be required.

`ad_integration_user` is used to verify `mssql_ad_sql_user_name` is configured correctly, which could be done by the user itself which is why I'm saying you can use `mssql_ad_sql_user_name` as the default value, and you're allowing it to log in to the SQL database, which is the below issue.

> 
> > > ad_integration_user is also required to create this user in SQL Server in https://github.com/linux-system-roles/mssql/blob/main/tasks/main.yml#L848 So it's good if the role ensures that is it is present.
> > 
> > 
> > Why are you creating the user account for `ad_integration_user`? If that account has the ability to add computers to the domain and create and modify user accounts, it should not be an account that you're logging in to the SQL database with. Again, least privileged access needs to be a core concept that is encouraged. You can give options to ignore LPA, but it shouldn't be the default or the only option.
> 
> Per the documentation in https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-active-directory-authentication?view=sql-server-ver16#createsqllogins or https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-ad-auth-adutil-tutorial?view=sql-server-ver16#create-active-directory-based-sql-server-logins-in-transact-sql I specifically verified with Microsoft developer that here the login with the username of ad_integration privileged user must be created, so `ad_integration_user`.

Those commands are to allow a user to log in to the SQL server. It's even labeled "Create Active Directory-based logins in Transact-SQL". My point is that you shouldn't be using the highly privileged `ad_integration_user`  to log in to your SQL server. Instead you should be using an account that should have that access. That access should be granted in the `mssql_post_input_sql_file` script. Those commands do need to be run in order to allow an AD user to log in, but they should target a user or group that is supposed to be allowed to log in, which is going to vary based on the company and is going to be much easier to manage if they're grouped together with the GRANT statements.

spetrosi:
> > > Why are you creating the user account for `ad_integration_user`? If that account has the ability to add computers to the domain and create and modify user accounts, it should not be an account that you're logging in to the SQL database with. Again, least privileged access needs to be a core concept that is encouraged. You can give options to ignore LPA, but it shouldn't be the default or the only option.
> > 
> > 
> > Per the documentation in https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-active-directory-authentication?view=sql-server-ver16#createsqllogins or https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-ad-auth-adutil-tutorial?view=sql-server-ver16#create-active-directory-based-sql-server-logins-in-transact-sql I specifically verified with Microsoft developer that here the login with the username of ad_integration privileged user must be created, so `ad_integration_user`.
> 
> Those commands are to allow a user to log in to the SQL server. It's even labeled "Create Active Directory-based logins in Transact-SQL". My point is that you shouldn't be using the highly privileged `ad_integration_user` to log in to your SQL server. Instead you should be using an account that should have that access. That access should be granted in the `mssql_post_input_sql_file` script. Those commands do need to be run in order to allow an AD user to log in, but they should target a user or group that is supposed to be allowed to log in, which is going to vary based on the company and is going to be much easier to manage if they're grouped together with the GRANT statements.

So there should be another variable with which users must provide a list of logins to create in SQL Server for this purpose, called e.g. `mssql_ad_based_logins`. And would it make sense to instead use `mssql_ad_sql_user_name` by default here (probably not, but I want to verify?

spetrosi:
> My choice would be no to `mysql_ad_based_logins`. Like I said, allowing someone to login should be grouped with their access (GRANT statements) to make it easier to ensure that everyone that can login has some form of access and everyone who has access can login. If you have a list of users/groups that can login in the role, then the GRANT statements still need to be in the post file and then you need to make sure the two completely separate lists have the same users/groups listed. It's not worth trying to make generating the GRANT statements part of the role because you need the databases created first. You'll end up adding complexity for very little return.

So should I rather remove this `Create the privileged account login` task to avoid security issues?

JasonN3:
> So should I rather remove this `Create the privileged account login` task to avoid security issues?

Yeah. That way there's no accounts created unexpectedly. If they don't create any users in the post file, they'll still have the `sa` account

spetrosi:
@JasonN3 please check the last commit where I removed creating the privileged account login. I would like to chat about it sometime next week. Could you add me on slack or email me on spetrosi@redhat.com? Thank you

JasonN3:
I put in 2 review comments that show up much earlier in this thread

defaults/main.yml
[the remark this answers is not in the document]
spetrosi: Changed to mssql_ad_join everywhere

spetrosi:
@JasonN3 I think I applied all your corrections. I decided to add another variable `mssql_ad_obtain_kerberos` to control whether to obtain kerberos or not. It is more transparent then deciding on whether users provide `ad_integration_user` or not.
I moved removing creating the privileged account login to #217 for clarity.
Please review.

tasks/main.yml
```diff
@@ -661,17 +661,33 @@
...
 
-    - name: Create keytab file
-      when: mssql_ad_keytab_file is none
+    - name: Join to realm {{ ad_integration_realm }}
+      include_role:
+        name: fedora.linux_system_roles.ad_integration
+      when: mssql_ad_join | bool
+
+    - name: Obtain Kerberos ticket
+      when: mssql_ad_obtain_kerberos
```
JasonN3, line 690: You still need a kerberos ticket if you want to be able to run any `adutil` commands later on. Another option:
If `ad_integration_user` is specified, run the same block and then also check if the user exists, try to create it if it doesn't, and set the SPN. Then, regardless of whether `ad_integration_user` was specified, get a kerberos ticket for `mssql_ad_sql_user_name` and then get the keytab using that account.

That way, you know a ticket was generated before trying to create a keytab file. This would also verify `mssql_ad_sql_password`

README.md
```diff
@@ -1402,6 +1438,51 @@ Type: `string`
...
+    mssql_accept_microsoft_cli_utilities_for_sql_server_eula: true
+    mssql_accept_microsoft_sql_server_standard_eula: true
+    mssql_version: 2022
+    mssql_password: "p@55w0rD"
+    mssql_edition: Evaluation
+    mssql_manage_firewall: true
+    mssql_ad_configure: true
+    mssql_ad_join: false
+    mssql_ad_obtain_kerberos: false
+    mssql_ad_sql_user_name: sqluser
```
JasonN3, line 1482: Is it worth renaming `mssql_ad_sql_user_name` to `mssql_ad_sql_user` to match `ad_integration_user`?

JasonN3 (changes requested):

README.md
richm: `mssql_ad_sql_user_name` is already part of the public API - we can't just get rid of it without breaking user playbooks

tasks/main.yml
spetrosi: My intent was to let users who like to obtain kerberos themselves can do this. By default, `mssql_ad_obtain_kerberos` is `true`, and I expect that the majority would keep it `true` and just provide `ad_integration_user`. I actually do not see an issue in obtaining kerberos ticket even if you already have it obtained, the role is idempotent and will just skip obtaining if the ticket is activated.

What's your intent? Maybe to let users choose for what username to obtain a kerberos ticket?

README.md
JasonN3: Ok. It was just a suggestion in order to standardize variable names.

As a way to allow renaming a variable, you could define it as:
```yaml
mssql_ad_sql_user: "{{ mssql_ad_sql_user_name | default('sqluser') }}"
```

tasks/main.yml
spetrosi: > Do you want to allow for situations where 3 accounts are used? 1 for AD joining, 1 for sqluser creationg, and sqluser.

I'd be happy to allow this if this is required. 

1. Account for AD joining is `ad_integration_user` - not applicable when `mssql_ad_join: false`
2. Account for sqluser creation is the one that obtains kerberos ticket. This would be either `ad_integration_user`, if provided, and if `ad_integration_user` is not provided, then `mssql_ad_sql_user_name`.
   Note that when `mssql_ad_join: true`, `ad_integration_user` will always be provided.  And if `ad_integration_user` is not defined, the `mssql_ad_kerberos_username` user must be pre-created in AD.

  The role can introduce another variable, like `mssql_ad_kerberos_username`, for the case when users want to join to AD and provide `ad_integration_user`, but want to use e.g. `mssql_ad_sql_user_name` or other user to obtain kerberos and create keytab.
3. `mssql_ad_sql_user_name`
Does it make sense?

README.md
spetrosi: Yeah, I can deprecate mssql_ad_sql_user_name and use mssql_ad_sql_user instead. I'll do it in a separate PR after the on-going PRs are merged.

tasks/main.yml
spetrosi: I ended up adding `mssql_ad_kerberos_user` and `mssql_ad_kerberos_password`, I think this is a needed flexibility. And I removed mssql_ad_obtain_kerberos, the role can ensure it for users, and in the case when users have a third AD user that they want to use for sqluser  and keytab creation, the can use the mssql_ad_kerberos_user variables.

JasonN3: I was actually saying 3 accounts might be a little excessive. I was only pointing out that having 3 accounts is the only situation I could see where having `mssql_ad_obtain_kerberos` not reflect whether `ad_integration_user` is set would be beneficial. If you want to have that flexibility in there, then I can completely agree with how each account would be used and the default values as you described them.

spetrosi: Haha, I didn't read the irony in your commit. But the resulted in a good thing.
IMO the logic of the default value of `mssql_ad_kerberos_user` is not exactly clear, maybe some users would want to set `mssql_ad_kerberos_user` to an actual username, whether it be the same as `ad_integration_user` or `mssql_ad_sql_user_name`, just to be explicit.

README.md
```diff
@@ -1283,30 +1283,53 @@ Optional: If you have already joined managed nodes to AD and you want to skip ru
...
+By default, the role uses the following logic:
 
-By default, when mssql_ad_obtain_kerberos is set to `true`, you must set `ad_integration_user` and `ad_integration_password` variables.
+1. If you define the `ad_integration_user` variable, the role obtains the kerberos ticket for the `ad_integration_user@ad_integration_realm` user with the `ad_integration_password` password.
 
-For an example playbook, see [Configuring AD integration with a pre-created keytab file](#configuring-ad-integration-without-joining-to-ad-and-without-obtaining-kerberos-ticket).
+2. If you don't define the `ad_integration_user` variable, the role obtains the kerberos ticket for `mssql_ad_sql_user_name@ad_integration_realm` user with the `mssql_ad_sql_password` password.
 
-Default: `true`
+You must provide this variable in the format <username>@{{ ad_integration_realm }}
```
JasonN3, line 1303: If they provide it in that format, it would result in <username>@<realm>@<realm>. In vars you set `__mssql_ad_kinit_user` to `{{ mssql_ad_kerberos_user }}@{{ ad_integration_realm | upper }}`. Either you need to have `mssql_ad_kerberos_user` specified as just a username or you need `__mssql_ad_kinit_user` to just be set to `mssql_ad_kerberos_user` and the default value of `mssql_ad_kerberos_user` to include the realm.

JasonN3 (changes requested):
A block is mislabeled and the requested format for `mssql_ad_kerberos_user` would result in realm being added twice, but otherwise, looks good to me. It would allow for situations where an administrative account was provided to create everything, where there is a separation of duty and everything has to be created beforehand, and where there is a separate account for each action.

tasks/main.yml
JasonN3: More flexibility is rarely a bad thing. Since it's already written, might as well keep it. It just means it can be used in more environments and it doesn't really complicate the code. It just adds a variable.

README.md
spetrosi: Good point, I first wrote this and then fixed the code to avoid this specific format, just forgot to remove the text.

tasks/main.yml
```diff
@@ -656,38 +656,23 @@
...
+        - mssql_ad_keytab_file none
 
     - name: Join to realm {{ ad_integration_realm }}
       include_role:
         name: fedora.linux_system_roles.ad_integration
       when: mssql_ad_join | bool
 
-    - name: Obtain Kerberos ticket
-      when: mssql_ad_obtain_kerberos
+    - name: Ensuer user in SQL server and create keytab file
```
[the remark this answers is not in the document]
spetrosi: It is now a bigger block, so it first obtains kerberos ticket, and the ensures the AD user (you're right it's not an SQL user) and it does create a keytab file.

tasks/main.yml
spetrosi: IMO it's good that the role always ensures that Kerberos ticket exists when a pre-generated `mssql_ad_keytab_file` file is provided. This is a simple step to automate, so I find the  `mssql_ad_obtain_kerberos` variable redundant.

spetrosi:
[citest]

README.md
spetrosi: Done in https://github.com/linux-system-roles/mssql/pull/220

PR #216, opened 2023-07-17
tests: tests_configure_ha_cluster add multi-node verification tasks

spetrosi:


spetrosi:
Related mssql-tests PR: idm-qe/mssql-tests/-/merge_requests/20

tests/tests_configure_ha_cluster.yml
```diff
@@ -172,15 +173,80 @@
           vars:
             mssql_ha_ag_cluster_type: external
 
+        - name: Assert that pcs runs properly when testing against >3 nodes
+          when: ansible_play_hosts_all | length >= 3
+          block:
+            - name: Wait for pcs to finish configuration
+              wait_for:
+                timeout: 200
```
richm, line 181: Is there some condition you can wait for instead?  is there some way to know if pcs is finished?

tomjelinek: I guess `crm_resource --wait` is what you are looking for. The command blocks as long as pacemaker is working on making the cluster status match the requested configuration (CIB).

richm: > Do you think it's a good idea to run `crm_resource --wait` in the role after ha_cluster role invocation? Maybe to run it directly inside ha_cluster role?

Yes - will require an rfe for the ha_cluster role - a parameter like `ha_cluster_wait_for_resource: true|false` or something like that?  By default it is `false` to preserve the existing ha_cluster role behavior.  If `true`, then the role would call `crm_resource --wait`

PR #217, opened 2023-07-18
fix!: AD - Remove creating the privileged account login

spetrosi:
Enhancement: For AD integration - remove the functionality to automatically create privileged account login in SQL Server.

Reason: This is required to ensure that the role does not create security leaks by silently creating accounts for users with admin permissions. Least privileged access and separation of duty are core security practices that the role should account.

Result: The role now does not create any Active Directory-based SQL Server logins, users are now responsible for doing this and are informed about this in README.md.


richm:
I get the security issue - but simply removing this will break customer playbooks/workflows if they are already using this and rely on it, regardless of how insecure it is.  Is there some way we can mitigate the security issue without breaking existing customer playbooks/workflows?

JasonN3:
> I get the security issue - but simply removing this will break customer playbooks/workflows if they are already using this and rely on it, regardless of how insecure it is. Is there some way we can mitigate the security issue without breaking existing customer playbooks/workflows?

A variable could be added to enable or disable the task, but if you enable it by default, you're causing the same security risk in new environments. If you disable it by default, existing customers will be affected the same way as removing it.

Maybe something could be added to the release notes that states if you were previously using the `ad_integration_user` to log in to your SQL servers, you will now need to add `CREATE LOGIN [{{ __mssql_ad_user }}] FROM WINDOWS;` to your `mssql_post_input_sql_file`

spetrosi:
I also do not see how we can fix this properly. It is dangerous to mark it as deprecated, because this is a valid security issue. Users for whom the current workflow would break should be grateful that there is no security leak now. Although, I understand that they can also be frustrated...
The role could print instructions on what users must accomplish to complete the configuration from the role. There are 2 post-config tasks currently.

spetrosi:
[citest]

richm:
lgtm - if this is a breaking change to the api, it will require a major version number change when it is released - so version 2.0.0

spetrosi:
I consider it a breaking change because of the new post-configuration task, but you would know better about semantic versioning. Would you consider it an API breaking change @richm.

richm:
> I consider it a breaking change because of the new post-configuration task, but you would know better about semantic versioning. Would you consider it an API breaking change @richm.

Yes, that, and removing `mssql_ad_netbios_name`

PR #218, opened 2023-07-24
feat: Add the ability to input T-SQL scripts content in addition to files

spetrosi:
Enhancement: Add the ability to add SQL scripts as a content directly.

Reason: Previously, users could save SQL script to a file and input it with `mssql_pre_input_sql_file` and `mssql_post_input_sql_file`. Sometimes for shorter scripts it may be an overkill to create a separate file, and it would be easier to provide the script content to the role directly.

Result: With this update, users can provide a T-SQL script directly with `mssql_pre_input_sql_content` and `mssql_post_input_sql_content` variables.

### DIFF
--- a/tasks/input_sql_files.yml
+++ b/tasks/input_sql_files.yml
@@ -36,6 +36,12 @@
     (__mssql_sqlcmd_login_cmd is none) or
     (__mssql_sqlcmd_login_cmd is not defined)
 
-- name: Input SQL files to MSSQL
+- name: Input SQL script files to MSSQL
   include_tasks: sqlcmd_input_file.yml
   loop: "{{ __mssql_sql_files_to_input }}"
+  when: __mssql_sql_files_to_input is defined
+
+- name: Input SQL script contents to MSSQL
+  include_tasks: sqlcmd_input_content.yml
+  loop: "{{ __mssql_sql_content_to_input }}"
+  when: __mssql_sql_content_to_input is defined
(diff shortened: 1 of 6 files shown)


tasks/input_sql_files.yml
```diff
@@ -36,6 +36,13 @@
     (__mssql_sqlcmd_login_cmd is none) or
     (__mssql_sqlcmd_login_cmd is not defined)
 
-- name: Input SQL files to MSSQL
+- name: Input SQL script files to MSSQL
   include_tasks: sqlcmd_input_file.yml
   loop: "{{ __mssql_sql_files_to_input }}"
+
+- name: Input SQL script contents to MSSQL
+  when: >-
```
richm, line 44: Why does the `when` check `mssql_pre_input_sql_content` and `mssql_post_input_sql_content` instead of just `__mssql_sql_content_to_input`?

spetrosi: That was a bug, if the loop_var is empty it won't be processed within the loop so this `when` was not required

richm (approved):

spetrosi:
[citest]

spetrosi:
[citest]

tasks/input_sql_files.yml
spetrosi: Hm no, after all this is required, what you was saying was correct

tasks/main.yml
```diff
@@ -415,6 +415,17 @@
   include_tasks: input_sql_files.yml
   when: mssql_pre_input_sql_file != []
 
+# Input content as list even when provided as sting
```
richm, line 418:
```suggestion
# Input content as list even when provided as string
```

PR #237, opened 2023-10-10
feat: Support mssql-server 2022 preview on RHEL 9

spetrosi:
Enhancement: Support mssql-server 2022 preview on RHEL 9

Reason: Microsoft added SQL Server 2022 preview for RHEL 9 at https://packages.microsoft.com/rhel/9/mssql-server-preview/

Result: You can install SQL Server 2022 on RHEL 9

Issue Tracker Tickets (Jira or BZ if any): https://issues.redhat.com/browse/RHELBU-2407


spetrosi:
[citest]

spetrosi:
> lgtm - should the rhel 9.4 tests be passing?

they must, there is some issue in cleanup. Also, Fedora tests are failing, due to /etc/yum.conf not present on Fedora, I'll fix support for Fedora in another PR to keep things clean.

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
Setting up mssql on RHEL 9.3 and 9.4 works for me locally. 
CI uses 9.4.0-20231010.15 from 10-Oct-2023 and I use the latest compose RHEL-9.4.0-20231011.18 from 11-Oct-2023
I'll test with what CI uses

spetrosi:
[citest bad]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

tasks/main.yml
```diff
@@ -196,6 +204,48 @@
...
 
+- name: Configure to run as a confined application with SELinux
+  when: __mssql_confined_supported | bool
+  block:
+    - name: Ensure that the system is in SELinux enforcing state
+      when: mssql_manage_selinux | bool
+      include_role:
+        name: fedora.linux_system_roles.selinux
+      vars:
+        selinux_state: "{{
```
richm, line 215:
```yaml
selinux_state: "{{ mssql_run_selinux_confined | ternary('enforcing', 'permissive') }}"
```

tasks/main.yml
```diff
@@ -196,6 +204,48 @@
...
+          'enforcing' if mssql_run_selinux_confined | bool
+          else 'permissive' }}"
+
+    - name: Get the current mode of SELinux
```
richm, line 219: Is it necessary to check that the selinux role set the correct mode?

defaults/main.yml
```diff
@@ -74,3 +74,6 @@ mssql_ad_kerberos_user: >-
 mssql_ad_kerberos_password: >-
   {{ ad_integration_password if ad_integration_user is defined
   else mssql_ad_sql_password }}
+
+mssql_manage_selinux: false
```
richm, line 78: I guess `mssql_manage_selinux` installs (or removes) the mssql selinux package, and `mssql_run_selinux_confined` says whether to run in Enforcing or Permissive if `mssql_manage_selinux` is set.  I guess it might be useful to run with selinux enabled but in Permissive mode for debugging purposes, but in general I would think if the user sets `mssql_manage_selinux: true` then they want to run with Enforcing.  If you are getting issues with selinux, or are anticipating issues, then having `mssql_run_selinux_confined` would be useful as a debugging option - perhaps then rename it as `__mssql_run_selinux_confined` so that users will know it is not for general use.

spetrosi: `mssql_run_selinux_confined` decides whether to install mssql-server-selinux package, having this package ensures that SQL Server runs as a confined application, this requires selinux in Enforcing.
`mssql_manage_selinux` decides whether to change selinux mode based on what `mssql_run_selinux_confined` is set to.
Amit had the same question in the email I sent yesterday. My reply was that mssql_manage_selinux is needed for those scenarios:

> * One scenario is when users have some other management platform in addition to Ansible that they manage selinux with, we do not want Ansible to conflict with it.
> * Other scenario is to prevent the role to change selinux mode silently without user's confirmation. Consider those who set their RHEL to selinux permissive, if then role changes their selinux back to enforcing their system might throw errors from many apps and it would be hard to troubleshoot what went wrong.

WDYT?

tasks/main.yml
spetrosi: Selinux role is run when `mssql_manage_selinux: true`, if `mssql_manage_selinux: false` the role gets the current mode and then verifies if the mode matches the chosen scenario confined vs non-confined

tasks/main.yml
spetrosi: Thanks, I forgot about ternary

defaults/main.yml
spetrosi: Also, now SQL Server 22 on RHEL 9 is in preview, GA is expected in 2024. Microsoft might implement some feedback in GA, e.g. implement checking for used ports SELinux type to match `mssql_port_t`, then mssql_manage_selinux would fix SELinux type for custom ports.

richm (approved):

spetrosi:
Let's wait with merging until reviews and comments in email thread are resolved.

spetrosi:
[citest]

spetrosi:
[citest bad]

spetrosi:
[citest bad]

PR #241, opened 2023-11-27
fix: Remove no_log: true where it is not required

### LINKED ISSUE #240: the output has been hidden due to the fact that 'no_log: true' was specified for this result
Encountered error while running playbook on TASK [microsoft.sql.server : Gather package facts]
Error msg: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": false}

Need help on this!

spetrosi:
Enhancement: Remove `no_log: true` where it is not required.

Reason: `no_log: true` was set on `package_facts` and `service_facts` modules to have cleaner logs because the large output of this task is not helpful for logs. However, users might have issues with Ansible not able to run these modules e.g. because it is not able to find python interpreter, such issues were hard to troubleshoot with a hidden output.

Result: Tasks that use `package_facts` and `service_facts` modules do not use `no_log: true`, hence they print output when running `ansible-playbook` with `-v`.

Fixes #240

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -106,7 +106,6 @@
 - name: Gather package facts
   package_facts:
     manager: auto
-  no_log: true
 
 - name: Set fact with the currently installed SQL Server version if any
   set_fact:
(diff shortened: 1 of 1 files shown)


spetrosi:
[citest]

PR #247, opened 2023-12-05
fix: Manage selinux context for custom storage directories

spetrosi:
Enhancement: On RHEL 9, manage SELinux context for custom storage paths

Reason: SQL Server runs on RHEL 9 as a SELinux-confined application and requires custom storage paths to have the correct SELinux context.

Result: When user sets `mssql_manage_selinux: true`, the role configures directories provided with `mssql_logdir` and `mssql_datadir` with a proper SELinux context.

### DIFF
--- a/tasks/configure_storage_paths.yml
+++ b/tasks/configure_storage_paths.yml
@@ -20,6 +20,22 @@
       else
       __mssql_storage_mode_default }}"
 
+- name: Append facts for the selinux role
+  vars:
+    selinux_fcontext:
+      - target: "{{ __mssql_storage_path }}"
+        setype: "{{ 'mssql_db_t' if __mssql_storage_setting == 'defaultlogdir'
+          else 'mssql_var_t' }}"
+        ftype: d
+        state: present
+    selinux_restore_dir:
+      - "{{ __mssql_storage_path }}"
+  set_fact:
+    selinux_fcontexts: "{{
+      (selinux_fcontexts | default([])) + selinux_fcontext }}"
+    selinux_restore_dirs: "{{
+      (selinux_restore_dirs | default([])) + selinux_restore_dir }}"
+
 - name: Configure the setting {{ __mssql_storage_setting }}
   include_tasks: mssql_conf_setting.yml
   vars:
--- a/tests/tasks/cleanup.yml
+++ b/tests/tasks/cleanup.yml
@@ -1,7 +1,6 @@
 ---
 - name: Gather package facts
   package_facts:
-    manager: auto
   no_log: true
 
 - name: Purge cluster configuration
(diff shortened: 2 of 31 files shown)


spetrosi:
[citest]

spetrosi:
[citest]

tasks/configure_storage_paths.yml
```diff
@@ -20,6 +20,23 @@
       else
       __mssql_storage_mode_default }}"
 
+- name: Append facts for the selinux role
+  vars:
+    selinux_fcontext:
+      - {
```
richm, line 26:
```suggestion
      - target: "{{ __mssql_storage_path }}"
        setype: "{{ 'mssql_db_t' if __mssql_storage_setting == 'defaultlogdir'
          else 'mssql_var_t' }}"
        ...
```
not sure why all of the selinux role examples mix json style `{ item: "value", ... }` with yaml style - should fix that in the selinux role to use only yaml style

spetrosi: I was thinking that this is a dictionary inside a list. I'll test it as you suggest.

spetrosi: It worked!

spetrosi:
[citest bad]

spetrosi:
[citest]

spetrosi:
[citest]

tests/tasks/cleanup.yml
```diff
@@ -63,11 +62,13 @@
...
     /etc/systemd/system/multi-user.target.wants/mssql-server.service
   register: __mssql_cleanup_remove
   changed_when: "'removed' in __mssql_cleanup_remove.stdout"
 
 # On SQL Server 2017, 2019 the service remains after removing RPMs
 - name: Stop the mssql-server service  # noqa command-instead-of-module
-  command: systemctl stop mssql-server  # noqa ignore-errors
-  ignore_errors: true  # noqa no-changed-when
+  shell: systemctl stop mssql-server || true
+  changed_when: false
```
richm, line 74: Can you use the `service:` module or the `systemd:` module instead of `shell: systemctl ...`?

spetrosi: mssql-server.service acts weirdly sometimes, and this is one of this. In the past, running this command was successful, but doing it with a `service` module failed because Ansible didn't find such a service. I'll check once again now.

spetrosi:
[citest]

spetrosi:
[citest]

tests/tasks/cleanup.yml
spetrosi: When stopping using the `sevrice` module, it fails with `Could not find the requested service mssql-server: host`. But this is required for a proper cleanup.

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

PR #253, opened 2024-01-15
fix:  Ensure selinux type for used ports, set default data and log dirmode

spetrosi:
Enhancement:

1. Ensure SELinux type for used ports
2. Set default mode for data and log storage directories 

Reason:

1. Custom TCP ports must have `mssql_port_t` SELinux type
3. `mssql_datadir_mode` and `mssql_logdir_mode` should have a default value for security

Result:

1. When mssql_manage_selinux is set to `true`, the role configures used ports with the `mssql_port_t` SELinux type
2. `mssql_datadir_mode` and `mssql_logdir_mode` variables have a default value of `'755'`

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,12 +413,9 @@ Type: `string`
 
 The permissions to be set for the `mssql_datadir` path in the format of the Ansible `file` module `mode` variable.
 
-Quote the mode like `'0700'` so Ansible parses it as a string to avoid conflicts with octal numbers.
+Quote the mode like `'755'` so Ansible parses it as a string to avoid conflicts with octal numbers.
 
-If mode is not specified and the destination directory *does not* exist, the role uses the default umask on the system when setting the mode.
-If mode is not specified and the destination directory *does* exist, the role uses the mode of the existing directory.
-
-Default: `null`
+Default: `'755'`
 
 Type: `string`
 
@@ -437,12 +434,9 @@ Type: `string`
 
 The permissions to be set for the `mssql_logdir` path in the format of the Ansible `file` module `mode` variable.
 
-Quote the mode like `'0700'` so Ansible parses it as a string to avoid conflicts with octal numbers.
+Quote the mode like `'755'` so Ansible parses it as a string to avoid conflicts with octal numbers.
 
-If mode is not specified and the destination directory *does not* exist, the role uses the default umask on the system when setting the mode.
-If mode is not specified and the destination directory *does* exist, the role uses the mode of the existing directory.
-
-Default: `null`
+Default: `'755'`
 
 Type: `string`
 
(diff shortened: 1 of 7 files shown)


richm:
doesn't this indicate that there is some sort of selinux problem with the way mssql-conf works?

richm:
We could use the selinux role to add the policy to allow that command.  https://github.com/linux-system-roles/selinux#selinux_ports

maybe something like this:
```yaml
selinux_ports:
  - ports: 1433
    proto: tcp
    setype: mssql_conf_t
    state: present
    local: true
```
?

richm (approved):

richm:
> @richm And I couldn't set port context to `mssql_conf_t` as you suggested.
> 
> ```
> [root@vm-10-0-187-18 ~]# semanage port -l | grep mssql
> mssql_port_t                   tcp      1433-1434
> mssql_port_t                   udp      1433-1434
> [root@vm-10-0-187-18 ~]# semanage port -a -t mssql_conf_t -p tcp 1433
> ValueError: Type mssql_conf_t is invalid, must be a port type
> ```
> 
> Could you get any other useful information on what the role can do to fix selinux issues from the ausearch output that I pasted in the comment above?

No, not sure - @bachradsusi is there a way to take the output of ausearch or audit2allow, and convert that to some sort of input for the selinux role?

bachradsusi:
> type=AVC msg=audit(1704936829.500:1731): avc:  denied  { name_connect } for  pid=61915 comm="python3" dest=1433 scontext=unconfined_u:unconfined_r:mssql_conf_t:s0-s0:c0.c1023 tcontext=system_u:object_r:mssql_port_t:s0 tclass=tcp_socket permissive=0

This problem can't be solved using `semanage`. There's missing rule which would allow 

`mssql_conf_t` process is not allowed to `name_connect` on `mssql_port_t`

`mssql_port_t` is not a standard port, it seems to be defined in `mssql-server-selinux` and therefore it should be fixed there.

It's also possible to ship local policy which would allow this. The policy could be generated using something like 

```
sesearch .... | audit2allow -M local_mssql
```

But it could be dangerous to do automatically and it's always necessary to review generated rules.

spetrosi:
[citest]

spetrosi:
Even though `ausearch` returns a lot of denials, the role itself works. I think that for now the role does all that it could to configure selinux, all other things should work out of the box. Microsoft doesn't describe any further steps, only install the `mssql-server-selinux` package and ensure that selinux is in enforcing mode.
Actually, the doc says to first install RPMs and then run `mssql-conf setup`, I think it's still better to be consistent with what MS docs say, so I'll probably revert this change here.

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

PR #260, opened 2024-02-21
fix: Running selinux unconfined is supported with permissive mode

spetrosi:
Enhancement: Add support for running SQL Server as a SELinux unconfined application with SELinux in enforcing mode on RHEL 9

Reason: This support was added by Microsoft

Result: Now you can run SQL Server as an unconfined application on RHEL 9 in the default enforcing SELinux mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,8 +201,9 @@ Whether to manage SELinux.
 
 When set to `true`, configures the following SELinux contexts and settings:
 
-1. SELinux `enforcing` or `permissive` mode based on the value of the [mssql_run_selinux_confined](#mssql_run_selinux_confined) variable.
+1. Ensures the `mssql_port_t` SELinux port type for [mssql_tcp_port](#mssql_tcp_port).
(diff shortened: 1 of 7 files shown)


spetrosi:
[citest]

richm:
I think your original indentation is correct - see https://github.com/linux-system-roles/ha_cluster/blob/main/tests/tests_cib_node_attributes.yml#L33

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
> I think your original indentation is correct - see https://github.com/linux-system-roles/ha_cluster/blob/main/tests/tests_cib_node_attributes.yml#L33

I reverted my change.
I remember ansible-lint complaining about my version of indentation (the one that I applied in this PR), so I changed to the original. But I do not understand why lists must be intended twice in this context. I was thinking that this was a bug in ansible-lint.

richm (approved):

spetrosi:
[citest]

PR #265, opened 2024-03-25
feat: support OpenSUSE/SLES

tisoft:
Enhancement:
Support deployment of MSSQL Server on OpenSUSE and SLES

Reason:
Suse systems are officially supported by the MSSQL RPM packages and this role only needs a few changes to also support them.

Result:
The role can be used to install and configure MSSQL  on OpenSUSE and SLES systems.

### DIFF
--- a/.collection/README.md
+++ b/.collection/README.md
@@ -2,13 +2,15 @@
 
 This collection contains a role for managing Microsoft SQL Server.
 
-## Currently supported distributions
+## Compatible OS
 
 * Red Hat Enterprise Linux 7 (RHEL 7+)
 * Red Hat Enterprise Linux 8 (RHEL 8+)
 * Red Hat Enterprise Linux 9 (RHEL 9+)
 * Fedora
 * RHEL derivatives such as CentOS
+* Suse Linux Enterprise Server 15 (SLES 15)
+* OpenSUSE 15
 
 ## Installing the collection
 
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -14,10 +14,18 @@ galaxy_info:
         - "9"
         - "8"
         - "7"
+    - name: SLES
+      versions:
+        - "15SP5"
+    - name: opensuse
+      versions:
+        - "15.5"
   galaxy_tags:
     - rhel
     - fedora
     - centos
+    - sles
+    - opensuse
     - mssql
     - sql
     - microsoft
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -188,10 +188,11 @@
   register: __mssql_gpg
   until: __mssql_gpg is success
 
-- name: Prepare for the upgrade
+- name: Prepare for the upgrade on RedHat
   when:
     - __mssql_current_version is defined
     - mssql_upgrade | bool
+    - ansible_pkg_mgr in ["yum", "dnf"]
   block:
     # This works only on systems that use yum or dnf
     - name: >-
@@ -223,15 +224,64 @@
         state: absent
       when: __mssql_curr_repo.stdout | length > 0
 
+- name: Prepare for the upgrade on Suse
+  when:
+    - __mssql_current_version is defined
+    - mssql_upgrade | bool
+    - ansible_pkg_mgr == "zypper"
+  block:
+    - name: >-
+        Update all packages from SQL Server repo
+        version {{ __mssql_current_version }}
+      package:
+        name: "*"
+        extra_args: >-
+          -r packages-microsoft-com-mssql-server-{{ __mssql_current_version }}
+        state: latest  # noqa package-latest
+
+    - name: Get zypper repositories  # noqa command-instead-of-module
+      shell: >-
+        set -euo pipefail;
+        zypper repos -E
+        | grep packages-microsoft-com-mssql-server-
+        | grep -v packages-microsoft-com-mssql-server-{{ mssql_version }}
+        | cut -d" " -f 4
+        || true
+      register: __mssql_zypper_curr_repo
+      changed_when: false
+      check_mode: false
+
+    - name: >-
+        Remove the current Microsoft SQL Server repository to upgrade to
+        {{ mssql_version }}
+      command: "zypper removerepo {{ __mssql_zypper_curr_repo.stdout }}"
+      when: __mssql_zypper_curr_repo.stdout | length > 0
+      changed_when: true
+
 - name: Configure the Microsoft SQL Server repo version {{ mssql_version }}
   yum_repository:
     name: packages-microsoft-com-mssql-server-{{ mssql_version | int }}
     description: Microsoft SQL Server {{ mssql_version }}
     baseurl: "{{ mssql_server_repository }}"
     gpgcheck: true
-  when: >-
-    (__mssql_server_packages not in ansible_facts.packages) or
-    (mssql_upgrade | bool)
+  when:
+    - ansible_pkg_mgr in ["yum", "dnf"]
+    - >-
+      (__mssql_server_packages not in ansible_facts.packages) or
+      (mssql_upgrade | bool)
+
+- name: Configure the Microsoft SQL Server repo version {{ mssql_version }} # noqa command-instead-of-module
+  command: >-
+    zypper addrepo -c -f --name "Microsoft SQL Server {{ mssql_version }}"
+    {{ mssql_server_repository }} packages-microsoft-com-mssql-server-{{ mssql_version | int }}
+  when:
+    - ansible_pkg_mgr == "zypper"
+    - >-
+      (__mssql_server_packages not in ansible_facts.packages) or
+      (mssql_upgrade | bool)
+  register: __zypper_addrepo_server_output
+  failed_when: not __zypper_addrepo_server_output.rc in [0, 4]
+  changed_when: __zypper_addrepo_server_output.rc != 4
 
 - name: Configure to run as a confined application with SELinux
   package:
@@ -268,6 +318,7 @@
     fi
   changed_when: false
   register: __mssql_errorlog
+  check_mode: false
 
 - name: Gather system services facts
   service_facts:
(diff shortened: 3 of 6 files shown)


tasks/main.yml
```diff
@@ -223,15 +225,65 @@
...
       when: __mssql_curr_repo.stdout | length > 0
 
+- name: Prepare for the upgrade
+  when:
+    - __mssql_current_version is defined
+    - mssql_upgrade | bool
+    - __mssql_current_version is defined
+    - ansible_pkg_mgr == "zypper"
+  block:
+    # This works only on systems that use yum or dnf
```
richm, line 235:
```suggestion
```

tasks/main.yml
```diff
@@ -223,15 +225,65 @@
         state: absent
       when: __mssql_curr_repo.stdout | length > 0
 
+- name: Prepare for the upgrade
+  when:
+    - __mssql_current_version is defined
+    - mssql_upgrade | bool
+    - __mssql_current_version is defined
```
richm, line 232:
```suggestion
```

tasks/main.yml
```diff
@@ -223,15 +225,65 @@
...
+        | cut -d" " -f 4
+        || true
+      register: __mssql_curr_repo
+      changed_when: false
+      check_mode: false
+
+    - name: >-
+        Remove the current Microsoft SQL Server repository to upgrade to
+        {{ mssql_version }}
+      community.general.zypper_repository:
```
richm, line 258: Is there no way to do this with the package module?  The problem with using third party modules like this is that the module is evaluated at **file load** time - which means **all** users on **all** platforms will need to have installed `community.general` or **the role will fail**.  So why doesn't the `when` clause prevent this?  Because `when` is evaluated at **file processing** time, which is after **file load** time.

If there is no way to do this with the `package` module or other Ansible built-in, then maybe you could put all of the zypper stuff in a file `tasks/suse_packaging.yml` or something like that, then in `tasks/main.yml` do something like
```yaml
- name: Handle Suse packaging
  include_tasks: suse_packaging.yml
  when: ansible_facts["distribution"] == "Suse"
```
In this case, the `include_tasks` and `when` will be processed at **file processing** time, which means only Suse users will need to install the 3rd party collection.

tasks/main.yml
```diff
@@ -192,6 +192,8 @@
   when:
     - __mssql_current_version is defined
     - mssql_upgrade | bool
+    - __mssql_current_version is defined
```
spetrosi, line 195:
```suggestion
```

tasks/main.yml
```diff
@@ -192,6 +192,8 @@
...
+    - ansible_pkg_mgr == "yum" or ansible_pkg_mgr == "dnf"
```
spetrosi, line 196:
```suggestion
    - ansible_pkg_mgr in ["yum", "dnf"]
```

tasks/main.yml
```diff
@@ -223,15 +225,63 @@
         state: absent
       when: __mssql_curr_repo.stdout | length > 0
 
+- name: Prepare for the upgrade
```
spetrosi, line 228:
```suggestion
- name: Prepare for the upgrade on Suse
```

spetrosi: You can also add `on Red Hat` in the above block.

tasks/main.yml
```diff
@@ -223,15 +225,63 @@
...
+
+    - name: Get zypper repositories  # noqa command-instead-of-module
+      shell: >-
+        set -euo pipefail;
+        zypper repos -E
+        | grep packages-microsoft-com-mssql-server-
+        | grep -v packages-microsoft-com-mssql-server-{{ mssql_version }}
+        | cut -d" " -f 4
+        || true
+      register: __mssql_curr_repo
```
spetrosi, line 251: It's better to use a different name for the variable. e.g. `__mssql_zypper_curr_repo`. Just in case if we need to use the content of __mssql_curr_repo somewhere else, and the content is different for dnf and zypper.

tasks/main.yml
spetrosi: Or maybe you can safely do this using `command` or `shell` if this is possible.

vars/main.yml
```diff
@@ -59,3 +59,4 @@ __mssql_ha_private_key_dest: /var/opt/mssql/data/{{ mssql_ha_cert_name }}.pvk
 
 __mssql_ad_kinit_user: >-
   {{ mssql_ad_kerberos_user }}@{{ ad_integration_realm | upper }}
+__mssql_configure_tuned: true
```
spetrosi, line 62: Please move this line up in the file after `__mssql_conf_cli`, To keep it close with other general config related variables.

tasks/main.yml
```diff
@@ -422,6 +481,14 @@
     description: Microsoft SQL Server Tools
     baseurl: "{{ mssql_client_repository }}"
     gpgcheck: true
+  when: ansible_pkg_mgr == "yum" or ansible_pkg_mgr == "dnf"
```
spetrosi, line 484:
```suggestion
  when: ansible_pkg_mgr in ["yum", "dnf"]
```

spetrosi (commented):
Thanks for your contribution!

spetrosi:
@tisoft any updates on this?

tisoft:
I'm currently on vacation, will continue with this next week, when I'm back at work. :)

vars/main.yml
tisoft: i have changed it to __mssql_tuned_supported and made a new overwritable default mssql_manage_tuned, so that also redhat users may disable tuned, if they like.

tasks/main.yml
tisoft: I have changed it to use zypper commands directly

defaults/main.yml
```diff
@@ -79,3 +79,4 @@ mssql_ad_kerberos_password: >-
 
 mssql_run_selinux_confined: "{{ __mssql_confined_supported | d(false) }}"
 mssql_manage_selinux: false
+mssql_manage_tuned: "{{ __mssql_tuned_supported | d(false) }}"
```
spetrosi, line 82: Please move this to `vars/main.yml` and rename it to `__mssql_manage_tuned`. `defaults/main.yml` is for variables that the role expects users might set, while `mssql_manage_tuned` is more like an internal variable. Prefixing a variable with two underscores shows that it is an internal variable.

tasks/main.yml
```diff
@@ -367,22 +416,30 @@
   package:
     name: tuned-profiles-mssql
     state: present
+  when:
+    - mssql_manage_tuned
```
spetrosi, line 420: Change to a one-line when expression: `  when: mssql_manage_tuned`

tasks/main.yml
```diff
@@ -367,22 +416,30 @@
...
 
 - name: Ensure that the tuned service is started and enabled
   service:
     name: tuned
     state: started
     enabled: true
+  when:
+    - mssql_manage_tuned
```
spetrosi, line 428: Change to a one-line when expression: `  when: mssql_manage_tuned`

tasks/main.yml
```diff
@@ -367,22 +416,30 @@
...
 
 - name: Get the active Tuned profiles
   command: tuned-adm active
   changed_when: false
   register: __mssql_tuned_active_profiles
+  when:
+    - mssql_manage_tuned
```
spetrosi, line 435: Change to a one-line when expression: `  when: mssql_manage_tuned`

tasks/main.yml
```diff
@@ -422,6 +479,14 @@
     description: Microsoft SQL Server Tools
     baseurl: "{{ mssql_client_repository }}"
     gpgcheck: true
+  when: ansible_pkg_mgr in ["yum", "dnf"]
+
+- name: Configure the Microsoft SQL Server Tools repository # noqa command-instead-of-module
+  command: 'zypper addrepo -c -f --name "Microsoft SQL Server Tools" {{ mssql_client_repository }} packages-microsoft-com-prod'
```
spetrosi, line 485:
```suggestion
  command: >-
    zypper addrepo -c -f --name "Microsoft SQL Server Tools" {{ mssql_client_repository }}
    packages-microsoft-com-prod
```
Quotes are not required, split a long line.

spetrosi: Thank you for removing dependency for community.general!

tasks/main.yml
```diff
@@ -422,6 +479,14 @@
...
+  when: ansible_pkg_mgr == "zypper"
+  register: __zypper_addrepo_output
```
spetrosi, line 487: I see you use the variable with the same name `__zypper_addrepo_output` twice. I'd recommend using a different variable each time for easier debugging. Thanks

spetrosi (commented):
Thanks for your changes, I am adding some more comments.

tasks/main.yml
richm: and add `| bool` - `when: mssql_manage_tuned | bool`
https://github.com/redhat-cop/automation-good-practices/tree/main/coding_style#ansible-guidelines
"Use the `| bool` filter when using bare variables (expressions consisting of just one variable reference without any operator) in `when`." and `changed_when` and `failed_when` and `assert: that`

meta/main.yml
```diff
@@ -5,19 +5,12 @@ galaxy_info:
   company: Red Hat, Inc.
   license: MIT
   min_ansible_version: "2.9"
-  platforms:
```
spetrosi, line 8: Why do you remove this? We need this for ci and generally for role best practices.

tisoft: I tried to find out, what to enter for sles/opensuse and found https://github.com/ansible/ansible/issues/82453 which states, that this is not used anymore by the ansible galaxy.

spetrosi: Our CI uses it to determine for what platforms to run tests.

.collection/README.md
```diff
@@ -9,6 +9,8 @@ This collection contains a role for managing Microsoft SQL Server.
 * Red Hat Enterprise Linux 9 (RHEL 9+)
 * Fedora
 * RHEL derivatives such as CentOS
+* Suse Linux Enterprise Server 15 (SLES 15)
```
spetrosi, line 12: Please rename the subtitle above from "Currently supported distributions" to "Compatible OS" to be more specific.

spetrosi: I am stealing this smart approach from https://github.com/linux-system-roles/ha_cluster/pull/188

meta/main.yml
tisoft: what should I put in there for suse? Or just leave suse out?

spetrosi: Let's put SUSE according to https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts/system/distribution.py#L68
```
    - name: SUSE
      versions:
        - "15"
```

spetrosi: Sorry, `SUSE` doesn't pass ansible-lint. And ansible-lint knows better, please use `opensuse` as per https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/schemas/meta.json#L1269

spetrosi:
[citest]

PR #290, opened 2024-08-06
chore: Fail on RHEL or CentOS 10 because it's not supported

spetrosi:
Enhancement: Fail on RHEL or CentOS 10 because it's not supported

Reason: Microsoft SQL Server does not support RHEL or CentOS 10 managed hosts yet


spetrosi:
[citest]

PR #301, opened 2024-09-19
feat: Use new adutil functionality to enable AES encryption

spetrosi:
Enhancement: Use new adutil functionality to enable AES encryption

Reason: Since adutil version 1.1.83, adutil supports enabling Kerberos AES 128 and AES 256 bit encryption when creating and modifying the user.

Result: The role automates enabling Kerberos AES 128 and AES 256 bit encryption, so the manual post-configuration task  is not required.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -853,7 +853,7 @@
         - name: Ensure adutil and krb5-workstation
           package:
             name:
-              - adutil
+              - adutil >= 1.1.143-1
               - krb5-workstation
             state: present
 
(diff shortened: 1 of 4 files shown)


spetrosi:
Converting to draft, since adutil that supports this is not public yet. Currently, this is being tested.

tasks/main.yml
```diff
@@ -907,6 +907,16 @@
...
           changed_when: true
 
+        - name: Ensure proper configuration on user {{ mssql_ad_sql_user }}
+          command: >-
+            adutil user modify
+            --name {{ mssql_ad_sql_user }}
+            --distname {{ mssql_ad_sql_user_dn | quote }}
+            --enableaessupport
+            --debug
+          when: __mssql_adutil_account_exists.stdout | bool
```
richm, line 917: Does `__mssql_adutil_account_exists.stdout | bool` evaluate to `false` if stdout is empty, or `true` otherwise?

spetrosi: Command `adutil account exists --name sqluser --accept-eula` returns `true` or `false`

spetrosi:
[citest]

PR #307, opened 2024-11-06
feat: Add mssql_tools_versions to allow installing different versions

spetrosi:
Enhancement: Add `mssql_tools_versions` variable to allow installing different versions of `mssql-tools` package.
Add `mssql_tls_self_sign` variable for specifying whether self-signed certificates are used, applicable when users set `mssql_tls_enable: true`.

Reason: Microsoft added a new version of the `mssql-tools` package - `mssql-tools18`. This package adds some new functionality, which is not backward-compatible with the previous versions, hence Microsoft added a new package `mssql-tools18` instead of just bumping the version of `mssql-tools`.
With `mssql-tools18`, when self-signed TLS certificates are used, it is required to set `-C` flag with `sqlcmd` to trust certificates.

Result: You can set `mssql_tools_versions` to install version 17, 18, or both version in parallel. By default, the role now installs the newer mssql-tools18.
With mssql-tools18, when you use self-signed TLS certificates, you must set `mssql_tls_self_sign: true` to make sqlcmd trust your certificates.

Issue Tracker Tickets (Jira or BZ if any): https://issues.redhat.com/browse/SYSROLES-166

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,8 @@ In the case when this variable contains multiple elements, the role instals both
 
 The role does not remove mssql-tools packages if you unset a version with this variable.
 
+If you use version 18 and [mssql_tls_enable](#mssql_tls_enable) with self-signed certificates, you must set [mssql_tls_self_sign](#mssql_tls_self_sign) to `true`.
+
 Default: `[18]`
 
 Type: `list`
@@ -615,6 +617,16 @@ Type: `string`
 #### mssql_tls_force
 
 Set to `true` to replace the existing certificate and private key files on host if they exist at `/etc/pki/tls/certs/` and `/etc/pki/tls/private/` respectively.
+
+Default: `false`
+
+Type: `bool`
+
+#### mssql_tls_self_sign
+
+Whether certificates that you use are self-signed or not.
+
+Based on this, the role decides whether to run `sqlcmd` with the `-C` argument to trust certificates.
 
 Default: `false`
 
--- a/README.md
+++ b/README.md
@@ -83,6 +83,26 @@ Default: `null`
 
 Type: `string`
 
+#### mssql_tools_versions
+
+List of versions of SQL Tools package (mssql-tools) to install and use `sqlcmd` from.
+
+The role uses `sqlcmd` for the following tasks:
+
+* Checking password for the sa user.
+* Running T-SQL scripts with variables [mssql_pre_input_sql_file and mssql_post_input_sql_file](#mssql_pre_input_sql_file-and-mssql_post_input_sql_file), [mssql_pre_input_sql_content and mssql_post_input_sql_content](#mssql_pre_input_sql_content-and-mssql_post_input_sql_content).
+
+By default, installs and uses the latest version 18.
+
+You can set this variable to `[17, 18]` to install two versions in parallel.
+In the case when this variable contains multiple elements, the role instals both but uses latest provided version.
+
+The role does not remove mssql-tools packages if you unset a version with this variable.
+
+Default: `[18]`
+
+Type: `list`
+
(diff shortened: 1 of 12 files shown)


spetrosi:
[citest]

README.md
```diff
@@ -83,6 +83,25 @@ Default: `null`
 
 Type: `string`
 
+#### mssql_tools_versions
+
+List of versions of SQL Tools package (mssql-tools) to install and use `sqlcmd` from.
+
+The role uses `sqlcmd` for two :
```
richm, line 90:
```suggestion
The role uses `sqlcmd` for the following:
```

README.md
```diff
@@ -83,6 +83,25 @@ Default: `null`
...
+1. Checking password for the sa user.
```
richm, line 91:
```suggestion
* Checking password for the sa user.
```

README.md
```diff
@@ -83,6 +83,25 @@ Default: `null`
...
+2. Running T-SQL scripts with variables [mssql_pre_input_sql_file and mssql_post_input_sql_file](#mssql_pre_input_sql_file-and-mssql_post_input_sql_file), [mssql_pre_input_sql_content and mssql_post_input_sql_content](#mssql_pre_input_sql_content-and-mssql_post_input_sql_content).
```
richm, line 92:
```suggestion
* Running T-SQL scripts with variables [mssql_pre_input_sql_file and mssql_post_input_sql_file](#mssql_pre_input_sql_file-and-mssql_post_input_sql_file), [mssql_pre_input_sql_content and mssql_post_input_sql_content](#mssql_pre_input_sql_content-and-mssql_post_input_sql_content).
```

defaults/main.yml
```diff
@@ -7,6 +7,7 @@ mssql_version: null
 mssql_upgrade: false
 mssql_password: null
 mssql_edition: null
+mssql_tools_versions: [17, 18]
```
richm, line 10: By default you install both versions?

vars/main.yml
```diff
@@ -12,7 +12,15 @@ __mssql_required_facts_subsets: "{{ ['!all', '!min'] +
   __mssql_required_facts }}"
 __mssql_server_packages: mssql-server
 __mssql_server_selinux_packages: mssql-server-selinux
-__mssql_client_packages: [mssql-tools, unixODBC-devel]
+__sqlcmd_ver: "{{
```
richm, line 15:
```suggestion
__sqlcmd_ver: "{{ mssql_tools_versions | sort | last }}"
```
This should work if `length == 1` or `length > 1` - it seems a little unnecessary but 1) the underlying python code will probably be faster without the if condition 2) this is simpler

One thing - is there a way to make `sort` do a numeric comparison?  It doesn't really matter until version `100` which will be sorted before `99`, or if they change the way they do versions

defaults/main.yml
spetrosi: oops, didn't update it after doing my test.

spetrosi:
[citest]

vars/main.yml
spetrosi: Right, one line is better.
I hope that they end on version 18... or make all future versions backward-compatible and move the current mssql-tools to an exceptional `mssql-tools17`. So that other packages can be simply updated. We can act accordingly.

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

PR #318, opened 2024-11-25
chore: Deprecate mssql_accept_microsoft_odbc_driver_17_for_sql_server_eula

spetrosi:
Enhancement: Deprecate this variable to use a variable without a version number instead. The role now can install versions 17 and 18.

Reason: After adding `mssql_tools_versions`, this role is able to install odbc driver for mssql_tools 17 and 18. This update changes the variable for specific version 17 to be general.

Result: When users use the deprecated variable `mssql_accept_microsoft_odbc_driver_17_for_sql_server_eula`, the role notifies them to use the new variable, but continues to execute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,14 @@ These variables apply to general SQL Server configuration.
 
 ### General Settings Variables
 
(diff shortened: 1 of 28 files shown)


spetrosi:
[citest]

PR #319, opened 2024-11-26
ci: Add job for publishing collection to Galaxy on new tag

spetrosi:


spetrosi:
[microsoft-sql-2.5.1.tar.gz](https://github.com/user-attachments/files/17919313/microsoft-sql-2.5.1.tar.gz)
Example collection artifact that I got from a test run

spetrosi:
@richm WDYT? Please review only the first commit. The second commit is to run tests here.
I will need to add this to our .github automation.

.github/workflows/changelog_to_tag.yml
```diff
@@ -2,11 +2,7 @@
 # yamllint disable rule:line-length
 name: Tag, release, and publish role based on CHANGELOG.md push
 on:  # yamllint disable-line rule:truthy
-  push:
-    branches:
-      - main
-    paths:
-      - CHANGELOG.md
+  pull_request:
```
richm, line 5: Why use `pull_request:` here?  I think you only want this job to run when there is a merge for `CHANGELOG.md`?

.github/workflows/changelog_to_tag.yml
```diff
@@ -68,6 +81,77 @@ jobs:
           fi
           echo "tagname=$_tagname" >> "$GITHUB_OUTPUT"
           echo "branch=$_branch" >> "$GITHUB_OUTPUT"
+          popd
+
+      - name: Build and publish the collection
```
richm, line 82: Some of this stuff is done by `release_collection.py` - is there some reason you cannot use `release_collection.py` here?

.github/workflows/changelog_to_tag.yml
richm: never mind

.github/workflows/changelog_to_tag.yml
spetrosi: release_collection.py required collection_release.yml, I thought it is too specific to be used outside linux-system-roles and tried to do the same but in less steps.

spetrosi: Looking at https://github.com/linux-system-roles/auto-maintenance/blob/main/collection_release.yml, Check the collection using `galaxy-importer` is a step that I miss here. I'll take a look at using this script once again.

richm: > release_collection.py required collection_release.yml, I thought it is too specific to be used outside linux-system-roles and tried to do the same but in less steps.

Ah, right, needs that file with a list of roles to checkout.  Hmm - not sure if you can use release_collection.py then

spetrosi: release_collection.py has too many things related to linux-system-roles collection hard coded. There is not much code to reuse, I prefer it like it is now

PR #336, opened 2025-02-21
Update replicate_db.j2

zhuatuzilo:
add auto create db if not exists


spetrosi:
I think this may cause troubles. E.g. if users do a typo in a database name in `mssql_ha_db_names`, the role will create and replicate an extra database.

README suggests to use `mssql_pre_input_sql_file` to pre-create a database if it doesn't exist. See https://github.com/linux-system-roles/mssql/blob/main/README.md?plain=1#L697-L700

The role should fail with an informative error message if a database provided in `mssql_ha_db_names` does not exist, saying that the database with this name doesn't exist and that you can create it e.g. with `mssql_pre_input_sql_file`.

zhuatuzilo:
you are right，maybe it's just my personal need，and i will change this in my local

PR #344, opened 2025-04-22
ci: skip storage scsi, nvme tests in github qemu ci

richm:
These tests are problematic in github qemu tests, and that
functionality (scsi, anyway) in the testing farm integration
tests.

Yes, we should have a way to provide tags on a per-role basis . . .


martinpitt:
This fell through the cracks. I pushed an initial refresh to the current workflow and tox-lsr workflow. C9 and C10 are fine, but F42 has a lot of failures. I started looking at [this one](https://github.com/linux-system-roles/mssql/actions/runs/14605228672/job/40972650309?pr=344#step:11:520), which seems quite shallow:

```
TASK [Remove related packages] *************************************************
"Problem: problem with installed package\n  - installed package mssql-server-selinux-16.0.4185.3-3.x86_64 requires mssql-server, but none of the providers can be installed\n
```

That's because it tries to remove `mssql-server`, but not `mssql-server-selinux` which is depending on it. This is another instance of "dnf5's `package:` does not remove reverse dependencies" which already plagued us in https://github.com/linux-system-roles/logging/pull/440.

I pushed that initial fix. That's not sufficient yet, but let's give that a try.

martinpitt:
Oh nice, that went much better than I thought! F42 is green. There is [one failure on C9](https://github.com/linux-system-roles/mssql/actions/runs/14833595400/job/41639981657?pr=344#step:12:2901), which didn't happen before, and it could be a flake. Retrying.

martinpitt (approved):
@richm Thanks! Can we please land this? Then I'll build on top of that for container/bootc support.

richm:
> @richm Thanks! Can we please land this? Then I'll build on top of that for container/bootc support.

@spetrosi usually handles mssql role changes

spetrosi (approved):

PR #346, opened 2025-05-05
feat: Support role on RHEL 10

martinpitt:
- [ ] Builds on top of #344

#344 shows that the role works on Fedora 42, so it ought to work on RHEL 10 as well. Let's try!


martinpitt:
Ah, PR #290 -- @spetrosi , is that still true? I.e. should we keep it disabled, or is the reason for it obsolete now?

martinpitt:
Nevermind -- https://packages.microsoft.com/rhel/ still doesn't have 10. https://packages.microsoft.com/fedora/ has 42, though.

PR #347, opened 2025-05-06
fix: Make service check work in container environments; plus two other bootc preparation fixes

martinpitt:
This is not nearly sufficient yet, but some initial steps which I would like to test and land in isolation -- these shouldn't break the "booted system" case.

### DIFF
--- a/.github/workflows/qemu-kvm-integration-tests.yml
+++ b/.github/workflows/qemu-kvm-integration-tests.yml
@@ -105,7 +105,7 @@ jobs:
           python3 -m pip install --upgrade pip
           sudo apt update
           sudo apt install -y --no-install-recommends git ansible-core genisoimage qemu-system-x86
-          pip3 install "git+https://github.com/linux-system-roles/tox-lsr@3.7.0"
+          pip3 install "git+https://github.com/linux-system-roles/tox-lsr@3.7.1"
 
       - name: Configure tox-lsr
         if: steps.check_platform.outputs.supported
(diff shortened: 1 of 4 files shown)


richm (approved):

martinpitt:
[citest]

martinpitt:
[c9s failed tests_input_sql_file_2022](https://github.com/linux-system-roles/mssql/actions/runs/14861444924/job/41727096649?pr=347#step:12:1717) with

> Failed to download metadata for repo 'resilientstorage': Cannot prepare internal mirrorlist: Status code: 503 for https://mirrors.centos.org/metalink?repo=centos-resilientstorage-9-stream&arch=x86_64&protocol=https,http

This smells transient, that URL works right now for me, and a local run succeeds.

[f42 failed differently in tests_selinux_enforcing_2022](https://github.com/linux-system-roles/mssql/actions/runs/14861444924/job/41727096664?pr=347#step:12:2900), and that does not look like a transient network error. However, it passes locally.

In the allegedly failed [C7 run](https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/tf_mssql-347_CentOS-7-latest-2.9_20250506-135452/artifacts/) all tests passed. :man_shrugging: 

So let's retry one round.

martinpitt:
[citest_bad]

martinpitt:
[citest_bad]

martinpitt:
[test_selinux_enforcing_2022 fails the same way on main](https://github.com/linux-system-roles/mssql/actions/runs/14835035907/job/41644380096), so this is indeed an unrelated flake. I'll look at it separately.

PR #349, opened 2025-05-13
feat: Support this role in container builds

martinpitt:
feat: Support this role in container builds

Feature: Support running the mssql role during container builds.

Reason: This is particularly useful for building bootc derivative OSes.

Result: These flags enable running the bootc container scenarios in CI,
which ensures that the role works in buildah build environment. This
allows us to officially support this role for image mode builds.

Fixes https://issues.redhat.com/browse/RHEL-89581

## Summary by Sourcery

Enable the mssql role to run reliably in container build environments by detecting systemd boot state and conditionally adjusting tasks, services, and tests to support non-booted scenarios.

New Features:
- Detect systemd boot status via a new __mssql_is_booted flag.
- Support executing role tasks in non-booted (container) environments.

Enhancements:
- Conditionally gate service starts, Tuned profile operations, SQL inputs, edition settings, and handlers based on booted state.
- Manage Tuned profiles in containers by directly updating /etc/tuned/active_profile instead of using tuned-adm.
- Relax initial configuration failure checks to tolerate container startup errors.

Documentation:
- Add a 'containerbuild' tag to galaxy metadata to document container support.

Tests:
- Separate edition and version verification logic for booted vs non-booted systems.
- Tag and skip booted-only tests when running in container builds.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -384,6 +384,11 @@
       register: __mssql_conf_setup
       no_log: true
       changed_when: true
+      # in non-booted mode this tries to start the server, which fails; it
+      # still configures things before, so tolerate that
+      failed_when: >-
+        (__mssql_is_booted and __mssql_conf_setup.rc != 0) or
+        (not __mssql_is_booted and __mssql_conf_setup.rc not in [0, 1])
       until: >-
         "Initial setup of Microsoft SQL Server failed."
         not in __mssql_conf_setup.stdout and
@@ -447,27 +452,32 @@
 - name: Ensure that the tuned service is started and enabled
   service:
     name: tuned
-    state: started
+    state: "{{ 'started' if __mssql_is_booted else omit }}"
     enabled: true
   when: __mssql_tuned_supported | bool
 
 - name: Get the active Tuned profiles
-  command: tuned-adm active
-  changed_when: false
-  register: __mssql_tuned_active_profiles
+  slurp:
+    path: /etc/tuned/active_profile
+  register: __mssql_tuned_active_profiles_b64
+  when: __mssql_tuned_supported | bool
+
+- name: Decode the active Tuned profiles
+  set_fact:
+    __mssql_tuned_active_profiles: "{{ __mssql_tuned_active_profiles_b64.content | b64decode | trim }}"
   when: __mssql_tuned_supported | bool
 
 # adding the mssql profile to the end of the list ensures
 # that it overrides conflicting settings in other profiles
-- name: Add mssql to the list of Tuned profiles
+- name: Add mssql to the list of Tuned profiles on booted systems
   when:
+    - __mssql_is_booted | bool
     - __mssql_tuned_supported | bool
-    - '"mssql" not in __mssql_tuned_active_profiles.stdout'
+    - '"mssql" not in __mssql_tuned_active_profiles'
   block:
     - name: Attempt to add mssql to the list of Tuned profiles
       command: >-
-        tuned-adm profile {{ __mssql_tuned_active_profiles.stdout
-        | regex_replace('^Current active profile: ', '') }} mssql
+        tuned-adm profile {{ __mssql_tuned_active_profiles }} mssql
       register: __mssql_tuned_adm_profile
       changed_when: __mssql_tuned_adm_profile.stderr | length == 0
       failed_when: false
@@ -488,13 +498,28 @@
 
     - name: Add the fixed mssql profile to the list of Tuned profiles
       command: >-
-        tuned-adm profile {{ __mssql_tuned_active_profiles.stdout |
-        regex_replace('^Current active profile: ', '') }} mssql
+        tuned-adm profile {{ __mssql_tuned_active_profiles }} mssql
       when: >-
         "Cannot find profile 'throughput-performance'" in
         __mssql_tuned_adm_profile.stderr
       register: __mssql_tuned_adm_profile
       changed_when: __mssql_tuned_adm_profile.stderr | length == 0
+
+# Once the above hack gets dropped, this can become the primary method of
+# changing the profile; restart tuned.service on booted systems then
+- name: Add mssql to the list of Tuned profiles on non-booted systems
+  when:
+    - not __mssql_is_booted | bool
+    - __mssql_tuned_supported | bool
+    - '"mssql" not in __mssql_tuned_active_profiles'
+  block:
+    - name: Attempt to add mssql to the list of Tuned profiles
+      copy:
+        dest: /etc/tuned/active_profile
+        content: "{{ __mssql_tuned_active_profiles }} mssql"
+        owner: root
+        group: root
+        mode: "0644"
 
 - name: Configure the Microsoft SQL Server Tools repository
   yum_repository:
@@ -654,12 +679,15 @@
         "TCP Provider: Error code 0x2749" not in __mssql_password_query.stdout
         and
         "TCP Provider: Error code 0x2749" not in __mssql_password_query.stderr
+      when: __mssql_is_booted | bool
 
     - name: Ensure that the mssql-server service is stopped
       service:
         name: mssql-server
         state: stopped
-      when: __mssql_password_query is failed
+      when:
+        - __mssql_is_booted | bool
+        - __mssql_password_query is failed
       notify: Restart the mssql-server service
 
     - name: Gather package facts
@@ -674,7 +702,7 @@
           shell: >-
             MSSQL_SA_PASSWORD={{ mssql_password | quote }}
             {{ __mssql_conf_cli }} --noprompt set-sa-password
-          when: __mssql_password_query is failed
+          when: (not __mssql_is_booted) or (__mssql_password_query is failed)
           notify: Restart the mssql-server service
           register: __mssql_set_ha_password
           no_log: true
@@ -700,7 +728,9 @@
       else
       [mssql_pre_input_sql_file] }}"
   include_tasks: input_sql_files.yml
-  when: mssql_pre_input_sql_file != []
+  when:
+    - __mssql_is_booted | bool
+    - mssql_pre_input_sql_file != []
 
 # Input content as list even when provided as a string
 - name: Pre-input SQL script contents to SQL Server
@@ -711,7 +741,9 @@
       else
       [mssql_pre_input_sql_content] }}"
   include_tasks: input_sql_files.yml
-  when: mssql_pre_input_sql_content != []
+  when:
+    - __mssql_is_booted | bool
+    - mssql_pre_input_sql_content != []
 
 - name: Set a new edition for MSSQL
   when:
@@ -731,12 +763,15 @@
       register: __mssql_edition_matches
       changed_when: false
       check_mode: false
+      when: __mssql_is_booted | bool
 
     - name: Ensure that the mssql-server service is stopped
       service:
         name: mssql-server
         state: stopped
-      when: not __mssql_edition_matches.stdout | bool
+      when:
+        - __mssql_is_booted | bool
+        - not __mssql_edition_matches.stdout | bool
       notify: Restart the mssql-server service
 
     - name: Gather package facts
@@ -751,7 +786,7 @@
             MSSQL_PID: "{{ mssql_edition }}"
           register: __mssql_conf_set_edition
           changed_when: '"The new edition" in __mssql_conf_set_edition.stdout'
-          when: not __mssql_edition_matches.stdout | bool
+          when: not __mssql_is_booted or not __mssql_edition_matches.stdout | bool
           notify: Restart the mssql-server service
           no_log: true
       rescue:
@@ -1196,8 +1231,10 @@
             name: mssql-server
             state: restarted
           # noqa no-handler
-          when: (__mssql_conf_set is changed) or
-            (__mssql_server_ha_packages_install | d({}) is changed)
+          when:
+            - __mssql_is_booted | bool
+            - (__mssql_conf_set is changed) or
+              (__mssql_server_ha_packages_install | d({}) is changed)
 
         - name: Remove certificates
           when: mssql_ha_reset_cert | bool
@@ -1364,9 +1401,10 @@
             name: mssql-server
             state: restarted
           # noqa no-handler
-          when: (__mssql_conf_set is changed) or
-            (__mssql_server_ha_packages_install | d({}) is changed)
-          register: __mssql_replica_restarted
+          when:
+            - __mssql_is_booted | bool
+            - (__mssql_conf_set is changed) or
+              (__mssql_server_ha_packages_install | d({}) is changed)
 
         - name: Remove certificate from SQL Server
           vars:
--- a/tasks/set_vars.yml
+++ b/tasks/set_vars.yml
@@ -16,3 +16,23 @@
   vars:
     __mssql_vars_file: "{{ role_path }}/vars/{{ item }}"
   when: __mssql_vars_file is file
+
+- name: Determine if system is booted with systemd
+  when: __mssql_is_booted is not defined
+  block:
+    - name: Run systemctl
+      # noqa command-instead-of-module
+      command: systemctl is-system-running
+      register: __is_system_running
+      changed_when: false
+      failed_when: false
+
+    - name: Require installed systemd
+      fail:
+        msg: "Error: This role requires systemd to be installed."
+      when: '"No such file or directory" in __is_system_running.msg | d("")'
+
+    - name: Set flag to indicate that systemd runtime operations are available
+      set_fact:
+        # see https://www.man7.org/linux/man-pages/man1/systemctl.1.html#:~:text=is-system-running%20output
+        __mssql_is_booted: "{{ __is_system_running.stdout != 'offline' }}"
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -4,3 +4,4 @@
   service:
     name: mssql-server
     state: restarted
+  when: __mssql_is_booted | bool
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -22,6 +22,7 @@ galaxy_info:
         - "15.5"
   galaxy_tags:
     - centos
+    - containerbuild
     - database
     - el7
     - el8
--- a/tasks/verify_password.yml
+++ b/tasks/verify_password.yml
@@ -4,6 +4,7 @@
   service:
     name: mssql-server
     state: started
+  when: __mssql_is_booted | bool
 
 - name: Check if a custom tcpport setting exist
   command: grep '^tcpport = ' {{ __mssql_conf_path }}
--- a/tests/tasks/cleanup.yml
+++ b/tests/tasks/cleanup.yml
@@ -79,6 +79,7 @@
 - name: Stop the mssql-server service  # noqa command-instead-of-module
   shell: systemctl stop mssql-server || true
   changed_when: false
+  when: __mssql_is_booted | bool
 
 - name: Get SELinux policy modules
   command: semodule -l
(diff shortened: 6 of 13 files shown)


tasks/main.yml
```diff
@@ -444,30 +449,37 @@
...
 
 # adding the mssql profile to the end of the list ensures
 # that it overrides conflicting settings in other profiles
-- name: Add mssql to the list of Tuned profiles
+- name: Add mssql to the list of Tuned profiles on booted systems
   when:
+    - __mssql_is_booted
     - __mssql_tuned_supported | bool
-    - '"mssql" not in __mssql_tuned_active_profiles.stdout'
+    - '"mssql" not in __mssql_tuned_active_profiles.content'
```
richm, line 478:
```suggestion
    - '"mssql" not in __mssql_tuned_active_profiles.content | b64decode'
```
hey - thanks, sourcery.ai!

tasks/main.yml
```diff
@@ -444,30 +449,37 @@
...
   block:
     - name: Attempt to add mssql to the list of Tuned profiles
       command: >-
-        tuned-adm profile {{ __mssql_tuned_active_profiles.stdout
-        | regex_replace('^Current active profile: ', '') }} mssql
+        tuned-adm profile {{ __mssql_tuned_active_profiles.content }} mssql
```
richm, line 482:
```suggestion
        tuned-adm profile {{ __mssql_tuned_active_profiles.content | b64decode }} mssql
```

tasks/main.yml
```diff
@@ -488,14 +500,29 @@
 
     - name: Add the fixed mssql profile to the list of Tuned profiles
       command: >-
-        tuned-adm profile {{ __mssql_tuned_active_profiles.stdout |
-        regex_replace('^Current active profile: ', '') }} mssql
+        tuned-adm profile {{ __mssql_tuned_active_profiles.content }} mssql
```
richm, line 503:
```suggestion
        tuned-adm profile {{ __mssql_tuned_active_profiles.content | b64decode }} mssql
```

tasks/main.yml
```diff
@@ -488,14 +500,29 @@
...
+- name: Add mssql to the list of Tuned profiles on non-booted systems
+  when:
+    - not __mssql_is_booted
+    - __mssql_tuned_supported | bool
+    - '"mssql" not in __mssql_tuned_active_profiles'
+  block:
+    - name: Attempt to add mssql to the list of Tuned profiles
+      copy:
+        dest: /etc/tuned/active_profile
+        content: "{{ __mssql_tuned_active_profiles }} mssql"
```
richm, line 521:
```suggestion
        content: "{{ __mssql_tuned_active_profiles.content | b64decode }} mssql"
```

tasks/set_vars.yml
```diff
@@ -16,3 +16,23 @@
...
+      # noqa command-instead-of-module
+      command: systemctl is-system-running
+      register: __is_system_running
+      changed_when: false
+      failed_when: false
+
+    - name: Require installed systemd
+      fail:
+        msg: "Error: This role requires systemd to be installed."
+      when: __is_system_running.msg is defined and "No such file or directory" in __is_system_running.msg
```
richm, line 33:
```suggestion
      when: '"No such file or directory" in __is_system_running.msg | d("")'
```

tasks/set_vars.yml
```diff
@@ -16,3 +16,23 @@
...
+
+    - name: Set flag to indicate that systemd runtime operations are available
+      set_fact:
+        # see https://www.man7.org/linux/man-pages/man1/systemctl.1.html#:~:text=is-system-running%20output
+        __mssql_is_booted: "{{ __is_system_running.stdout not in ['offline', 'degraded'] }}"
```
richm, line 38: I know that we have used this idiom in other roles - but we have also seen cases e.g. logging role where the system was clearly booted and mostly running, but `systemctl is-system-running` returned `degraded` because journald was not running.  Have you seen cases where `systemctl is-system-running` returns `degraded` during bootc build?

tasks/main.yml
```diff
@@ -625,7 +652,7 @@
       set_fact:
         __sqlcmd_cli: "{{ __sqlcmd_cli + ' -C' }}"
 
-- name: Set a new password for the MSSQL sa user
+- name: Set a new password for the MSSQL as user
```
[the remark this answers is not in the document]
martinpitt: Ah right, `sa` is an actual user. I misread it as a typo for "as". I reverted this change.

tasks/main.yml
```diff
@@ -384,6 +384,11 @@
       register: __mssql_conf_setup
       no_log: true
       changed_when: true
+      # in non-booted mode this tries to start the server, which fails; it
+      # still configures things before, so tolerate that
+      failed_when: >-
+        (__mssql_is_booted and __mssql_conf_setup.rc != 0) or
+        (not __mssql_is_booted and not __mssql_conf_setup.rc in [0, 1])
```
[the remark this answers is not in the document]
martinpitt: done.

tasks/main.yml
```diff
@@ -444,30 +449,37 @@
...
+    state: started
+  when:
+    - __mssql_is_booted
+    - __mssql_tuned_supported | bool
+
 - name: Get the active Tuned profiles
-  command: tuned-adm active
-  changed_when: false
+  slurp:
+    path: /etc/tuned/active_profile
```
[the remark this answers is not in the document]
martinpitt: Oops, fixed.

tasks/main.yml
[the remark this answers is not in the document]
martinpitt: See above, reverted -- I initially misread 'sa' as a typo.

tasks/main.yml
martinpitt: Right, all fixed.

tasks/main.yml
martinpitt: I centralized that into a new fact, as there are too many places which use this value.

tasks/set_vars.yml
```diff
@@ -16,3 +16,23 @@
   vars:
     __mssql_vars_file: "{{ role_path }}/vars/{{ item }}"
   when: __mssql_vars_file is file
+
+- name: Determine if system is booted with systemd
+  when: not __mssql_is_booted is defined
```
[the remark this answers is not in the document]
martinpitt: We use both styles in our code, "not ... is defined" more often. But sure, "is not defined" is a bit more pleasant to read.

tasks/set_vars.yml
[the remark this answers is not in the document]
martinpitt: Nope, this totally isn't how this works.

martinpitt: Ah nice, thanks! I'll apply that to the existing roles together with the "degraded" fix.

tasks/set_vars.yml
martinpitt: No, I haven't. Indeed you are right, "degraded" here is nonsense :see_no_evil: Maybe I meant "unknown"? But anyway, let's only consider "offline" here. Fixed, and I'll send PRs for the three roles which already have that. Thanks for spotting!

tasks/set_vars.yml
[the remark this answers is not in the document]
martinpitt: No no no, please stop hallucinating. In this case it looks like `"msg": "[Errno 2] No such file or directory: b'systemdctl'", "rc": 2, "stderr": ""`.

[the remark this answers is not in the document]
martinpitt: This doesn't work either. `is-system-running` exits with nonzero for e.g. "degraded".

tasks/set_vars.yml
martinpitt: Sent https://github.com/linux-system-roles/tlog/pull/166 https://github.com/linux-system-roles/logging/pull/449 https://github.com/linux-system-roles/cockpit/pull/217

martinpitt:
@sourcery-ai review

martinpitt:
>     * Consolidate repeated `__mssql_is_booted` when conditions by grouping related boot-dependent tasks into blocks or separate includes to reduce boilerplate and improve maintainability.

Matter of taste, let's see what the human reviewers think.

>     * Use `ansible.builtin.stat` or the `lookup('file')` plugin instead of `slurp` + `b64decode` to read `/etc/tuned/active_profile` for simpler and clearer logic.

lookup('file') only reads local files from the controller, but I need the file from the target machine. `stat` does not read file contens.

>     * Enhance the `__mssql_is_booted` detection to handle additional `systemctl is-system-running` outputs (e.g., `degraded`, `maintenance`) rather than only checking for `offline` to make boot detection more robust.

Nope :grin:

martinpitt:
[remaining failure](https://github.com/linux-system-roles/mssql/actions/runs/15013106773/job/42188565701?pr=349#step:12:2924) is the known intermittent SELinux rejection.

This is ready for review now. Sorry @richm for the noise -- your review was requested automatically yesterday already. That was draft on purpose (I mostly just pushed what I had on EOD to get initial test feedback)

martinpitt:
[citest]

martinpitt:
Whoops, back to draft, there [is a problem](ehttps://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/tf_mssql-349_CentOS-Stream-8-2.9_20250514-083148/artifacts/tests_input_sql_file_2022-ANSIBLE-2.9-test_playbooks_parallel-FAIL.log):

```
task path: /tmp/collections-w6w/ansible_collections/fedora/linux_system_roles/tests/mssql/tasks/cleanup.yml:26
Wednesday 14 May 2025  04:58:00 -0400 (0:00:00.048)       0:05:25.065 ********* 
fatal: [managed-node3]: FAILED! => {}

MSG:

The conditional check 'mssql_manage_selinux | d(false) | bool' failed. The error was: error while evaluating conditional (mssql_manage_selinux | d(false) | bool): {{ mssql_run_selinux_confined }}: 'mssql_run_selinux_confined' is undefined

```

spetrosi:
[citest]

richm:
> mssql_manage_selinux: "{{ mssql_run_selinux_confined }}"

> but doesn't define mssql_run_selinux_confined. That should be fine -- that's what the | d(false) is for? Other tests do exactly the same but don't fail on that, and later Ansible versions work too. 🤔

It depends.  When you evaluate a variable which has been set in a `vars:` context, it is evaluated with lazy evaluation - that is, as long as `mssql_run_selinux_confined` is defined before `mssql_manage_selinux` is used, it should be fine.  However, variables defined with `set_fact` are evaluated immediately.  And ansible 2.9 did this slightly differently than later ansible, which is why we see this problem with 2.9.

I think the fix is to ensure that `mssql_run_selinux_confined` is always defined before using `mssql_manage_selinux`

martinpitt:
I tried this:

```diff
--- tests/tasks/cleanup.yml
+++ tests/tasks/cleanup.yml
@@ -23,12 +23,19 @@
     name: fedora.linux_system_roles.firewall
   when: ansible_facts.packages.firewalld is defined
 
-- name: Purge selinux configuration
+# HACK: work around a bug with older Ansible that fails if
+# mssql_run_selinux_confined is not defined -- some tests set
+# mssql_manage_selinux to that
+- name: Ensure mssql_run_selinux_confined is defined
   vars:
-    selinux_all_purge: true
-  include_role:
-    name: fedora.linux_system_roles.selinux
-  when: mssql_manage_selinux | d(false) | bool
+    mssql_run_selinux_confined: "{{ mssql_run_selinux_confined | default(false) }}"
+  block:
+    - name: Purge selinux configuration
+      vars:
+        selinux_all_purge: true
+      include_role:
+        name: fedora.linux_system_roles.selinux
+      when: mssql_manage_selinux | d(false) | bool
 
 - name: Leave realm
   command: realm leave
```

but that also fails:

> fatal: [/var/home/martin/.cache/linux-system-roles/centos-8.qcow2]: FAILED! => {"msg": "The conditional check 'mssql_manage_selinux | d(false) | bool' failed. The error was: An unhandled exception occurred while templating '{{ mssql_run_selinux_confined }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while templating ... [looong repetition] ... 
> original message: recursive loop detected in template string: {{ mssql_run_selinux_confined | default(false) }}

richm:
> I tried this:
> 
> ```diff
> --- tests/tasks/cleanup.yml
> +++ tests/tasks/cleanup.yml
> @@ -23,12 +23,19 @@
>      name: fedora.linux_system_roles.firewall
>    when: ansible_facts.packages.firewalld is defined
>  
> -- name: Purge selinux configuration
> +# HACK: work around a bug with older Ansible that fails if
> +# mssql_run_selinux_confined is not defined -- some tests set
> +# mssql_manage_selinux to that
> +- name: Ensure mssql_run_selinux_confined is defined
>    vars:
> -    selinux_all_purge: true
> -  include_role:
> -    name: fedora.linux_system_roles.selinux
> -  when: mssql_manage_selinux | d(false) | bool
> +    mssql_run_selinux_confined: "{{ mssql_run_selinux_confined | default(false) }}"
> +  block:
> +    - name: Purge selinux configuration
> +      vars:
> +        selinux_all_purge: true
> +      include_role:
> +        name: fedora.linux_system_roles.selinux
> +      when: mssql_manage_selinux | d(false) | bool
>  
>  - name: Leave realm
>    command: realm leave
> ```
> 
> but that also fails:
> 
> > fatal: [/var/home/martin/.cache/linux-system-roles/centos-8.qcow2]: FAILED! => {"msg": "The conditional check 'mssql_manage_selinux | d(false) | bool' failed. The error was: An unhandled exception occurred while templating '{{ mssql_run_selinux_confined }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while templating ... [looong repetition] ...
> > original message: recursive loop detected in template string: {{ mssql_run_selinux_confined | default(false) }}

Which test is this?

martinpitt:
See [above](https://github.com/linux-system-roles/mssql/pull/349#issuecomment-2879454290):
```
tox -e qemu-ansible-2.9 -- --image-name centos-8 --log-level=debug tests/tests_input_sql_file_2022.yml
```

I am trying something else now. Less local, but much simpler and robust.

martinpitt:
Fix pushed to #350 and also here.

martinpitt:
[citest]

tests/tests_idempotency_2022.yml
```diff
@@ -8,7 +8,7 @@
     mssql_accept_microsoft_cli_utilities_for_sql_server_eula: true
     mssql_accept_microsoft_sql_server_standard_eula: true
     mssql_version: 2022
-    mssql_manage_selinux: "{{ mssql_run_selinux_confined }}"
+    mssql_manage_selinux: "{{ mssql_run_selinux_confined | d(false) }}"
```
richm, line 11: Yes, this will fix the error, but it might not be testing what you want to test.  I'll defer to @spetrosi but I think the intention here is to test the manage selinux functionality if mssql has the ability be run confined, which must be dynamically determined later.

martinpitt: OK -- let's handle that in #350 then. On my next push I'll remove that commit here. TBH I'm very close to reverting the hack and just going back to running in ascending order everywhere, and going back to sweeping this problem under the rug -- it's just too damn hard, already cost me a week.

martinpitt: moved to #350, dropped the commit here.

tasks/main.yml
```diff
@@ -444,30 +449,42 @@
...
+    - __mssql_is_booted
+    - __mssql_tuned_supported | bool
+
 - name: Get the active Tuned profiles
-  command: tuned-adm active
-  changed_when: false
-  register: __mssql_tuned_active_profiles
+  slurp:
+    path: /etc/tuned/active_profile
+  register: __mssql_tuned_active_profiles_b64
```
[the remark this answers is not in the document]
martinpitt: Multiple profiles are actually being used and tested by mssql. They just appear as space separated names in the config file.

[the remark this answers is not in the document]
martinpitt: Yes, the decoding happens in the "Decode the active Tuned profiles" step 5 lines below.

tasks/main.yml
```diff
@@ -444,30 +449,42 @@
...
-    state: started
     enabled: true
   when: __mssql_tuned_supported | bool
 
+- name: Ensure that the tuned service is started
+  service:
+    name: tuned
+    state: started
+  when:
+    - __mssql_is_booted
```
richm, line 463:
```suggestion
    - __mssql_is_booted | bool
```
Not sure if this is needed anymore - it used to be required - but it also makes it consistent with the line below

martinpitt: I didn't `| bool`ify that flag anywhere else, nor in any other role. There's no need to, it's already a boolean, and if it isn't, we messed up and want to know (i.e. tests fail).

`__mssql_tuned_supported` is suspect indeed: there is *no* default value in vars/main.yml or anywhere else, just these two special cases:
```
vars/RedHat.yml:__mssql_tuned_supported: true
vars/Suse.yml:__mssql_tuned_supported: false
```

So for e.g. Fedora this comes out as `undefined | bool`, which I assume is `false` instead of an error. I'd prefer to clean this up and add `__mssql_tuned_supported: false` to vars/main.yml.

I sent PR #354 to clean this up.

richm: > I didn't `| bool`ify that flag anywhere else, nor in any other role. There's no need to, it's already a boolean, and if it isn't, we messed up and want to know (i.e. tests fail).

The `| bool` (or similar) used to be required because of the Ansible bare variable issue - https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.8.html#bare-variables-in-conditionals - which is still a thing in ansible 2.9 - so we *might* still need it when running on 2.9 - the reason why ansible-lint, ansible-test, etc. do not warn is because they don't care about ansible 2.15 or earlier.

> 
> `__mssql_tuned_supported` is suspect indeed: there is _no_ default value in vars/main.yml or anywhere else, just these two special cases:
> 
> ```
> vars/RedHat.yml:__mssql_tuned_supported: true
> vars/Suse.yml:__mssql_tuned_supported: false
> ```
> 
> So for e.g. Fedora this comes out as `undefined | bool`, which I assume is `false` instead of an error.

It is defined on Fedora via the RedHat.yml vars file.  Because of the way this logic works - https://github.com/linux-system-roles/mssql/blob/main/tasks/set_vars.yml#L9 - `ansible_os_family == "RedHat"` on Fedora (and CentOS, Rocky, Alma, etc.)

>  I'd prefer to clean this up and add `__mssql_tuned_supported: false` to vars/main.yml.

And that's fine, but it won't do anything in this case.

> 
> I sent PR #354 to clean this up.

martinpitt: OK -- didn't seem to break the C7 tests in the other projects, but I just added `| bool` everywhere.

tasks/main.yml
```diff
@@ -488,14 +505,29 @@
...
+    - __mssql_tuned_supported | bool
+    - '"mssql" not in __mssql_tuned_active_profiles'
+  block:
+    - name: Attempt to add mssql to the list of Tuned profiles
+      copy:
+        dest: /etc/tuned/active_profile
+        content: "{{ __mssql_tuned_active_profiles }} mssql"
+        owner: root
+        group: root
+        mode: "0644"
```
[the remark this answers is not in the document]
martinpitt: No, because in non-booted mode, tuned isn't running.

tasks/main.yml
```diff
@@ -1364,9 +1401,10 @@
             name: mssql-server
             state: restarted
           # noqa no-handler
-          when: (__mssql_conf_set is changed) or
-            (__mssql_server_ha_packages_install | d({}) is changed)
-          register: __mssql_replica_restarted
```
martinpitt, line 1369: FTR, I dropped this because nothing evaluated `__mssql_replica_restarted`.

spetrosi (approved):

PR #350, opened 2025-05-14
tests: Add workaround for Ansible 2.9 SELinux cleanup failure

martinpitt:
The "conditionalize SELinux cleanup" hack in commit e1c8d1fdf973f that fixed a regression with different test order caused a regression itself: On older Ansible (observed on 2.9) it fails some tests with:

> The conditional check 'mssql_manage_selinux | d(false) | bool' failed.
> The error was: error while evaluating conditional
> (mssql_manage_selinux | d(false) | bool): {{ mssql_run_selinux_confined }}:
> 'mssql_run_selinux_confined' is undefined

So the `d()` default does not apply to the actual definition of the variable in these versions.

Avoid that by adding a default to ensure that the value isn't undefined.

----

See https://github.com/linux-system-roles/mssql/pull/349#issuecomment-2879454290 . Fixes [this failure](https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/tf_mssql-349_CentOS-Stream-8-2.9_20250514-083148/artifacts/tests_input_sql_file_2022-ANSIBLE-2.9-test_playbooks_parallel-FAIL.log) in the CentOS 8 TF tests. We missed that as #349 did not run `[citest]`, sorry about that!

## Summary by Sourcery

Work around Ansible 2.9 SELinux cleanup regression by defaulting the selinux management variable in test playbooks.

Bug Fixes:
- Ensure mssql_manage_selinux is always defined by adding a default(false) fallback to mssql_run_selinux_confined in test variables.

Tests:
- Update all 2022 test playbooks (idempotency, input SQL file, password, TCP firewall, TLS) to apply the default(false) filter for mssql_manage_selinux.

## Summary by Sourcery

Provide a fallback default for the SELinux management variable in test playbooks to avoid undefined-variable errors on Ansible 2.9.

Bug Fixes:
- Ensure mssql_manage_selinux is always defined by using a default(false) filter on mssql_run_selinux_confined in test variables.

Tests:
- Update all 2022 test playbooks (idempotency, input SQL file, password, TCP firewall, TLS) to apply the default(false) fallback for mssql_manage_selinux.

### DIFF
--- a/tests/tests_idempotency_2022.yml
+++ b/tests/tests_idempotency_2022.yml
@@ -8,7 +8,7 @@
     mssql_accept_microsoft_cli_utilities_for_sql_server_eula: true
     mssql_accept_microsoft_sql_server_standard_eula: true
     mssql_version: 2022
-    mssql_manage_selinux: "{{ mssql_run_selinux_confined }}"
+    mssql_manage_selinux: "{{ mssql_run_selinux_confined | d(false) }}"
     __mssql_gather_facts_no_log: true
   tasks:
     - name: Run test in a block to clean up in always
--- a/tests/tests_input_sql_file_2022.yml
+++ b/tests/tests_input_sql_file_2022.yml
@@ -8,7 +8,7 @@
     mssql_accept_microsoft_sql_server_standard_eula: true
     mssql_debug: true
     mssql_version: 2022
-    mssql_manage_selinux: "{{ mssql_run_selinux_confined }}"
+    mssql_manage_selinux: "{{ mssql_run_selinux_confined | d(false) }}"
     __mssql_gather_facts_no_log: true
   tasks:
     - name: Run test in a block to clean up in always
--- a/tests/tests_password_2022.yml
+++ b/tests/tests_password_2022.yml
@@ -7,7 +7,7 @@
     mssql_accept_microsoft_cli_utilities_for_sql_server_eula: true
     mssql_accept_microsoft_sql_server_standard_eula: true
     mssql_version: 2022
-    mssql_manage_selinux: "{{ mssql_run_selinux_confined }}"
+    mssql_manage_selinux: "{{ mssql_run_selinux_confined | d(false) }}"
     __mssql_gather_facts_no_log: true
   tasks:
     - name: Run test in a block to clean up in always
--- a/tests/tests_tcp_firewall_2022.yml
+++ b/tests/tests_tcp_firewall_2022.yml
@@ -8,7 +8,7 @@
     mssql_accept_microsoft_sql_server_standard_eula: true
     mssql_manage_firewall: true
     mssql_version: 2022
-    mssql_manage_selinux: "{{ mssql_run_selinux_confined }}"
+    mssql_manage_selinux: "{{ mssql_run_selinux_confined | d(false) }}"
     __mssql_gather_facts_no_log: true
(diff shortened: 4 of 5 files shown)


martinpitt:
[citest]

martinpitt:
That tests_selinux_enforcing failure is really killing us here :cry: But at least this fixes the RHEL/CentOS 8 regression.

It's still shown as [failed](https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/tf_mssql-350_CentOS-Stream-8-2.9_20250514-160602/artifacts/), but why exactly? all the tests are passed.

tests/tests_idempotency_2022.yml
```diff
@@ -8,7 +8,7 @@
     mssql_accept_microsoft_cli_utilities_for_sql_server_eula: true
     mssql_accept_microsoft_sql_server_standard_eula: true
     mssql_version: 2022
-    mssql_manage_selinux: "{{ mssql_run_selinux_confined }}"
+    mssql_manage_selinux: "{{ mssql_run_selinux_confined | d(false) }}"
```
martinpitt, line 11: Quoting @richm in https://github.com/linux-system-roles/mssql/pull/349#discussion_r2089332801:

> Yes, this will fix the error, but it might not be testing what you want to test. I'll defer to @spetrosi but I think the intention here is to test the manage selinux functionality if mssql has the ability be run confined, which must be dynamically determined later.

So this might be just wrong.

If it is, then :white_flag: -- We have the option to instead just revert that  bit from commit e1c8d1fdf and instead go back to just running tests in ascending order everywhere. Of course that will sweep the problem under the rug again, but I've already spent a week on this -- this is becoming too expensive. Do you happen to know a good contact for debugging deep SELinux issues who might help with this?

richm:
> That tests_selinux_enforcing failure is really killing us here 😢 But at least this fixes the RHEL/CentOS 8 regression.
> 
> It's still shown as [failed](https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/tf_mssql-350_CentOS-Stream-8-2.9_20250514-160602/artifacts/), but why exactly? all the tests are passed.

Good question.  First, go to https://github.com/linux-system-roles/mssql/actions to find the action and the run for this PR, which in this case is https://github.com/linux-system-roles/mssql/actions/runs/15025599646/job/42225475633.  Then, click on the link printed here:

```
Run sclorg/testing-farm-as-github-action@v4
Notice: Testing Farm logs: https://artifacts.osci.redhat.com/testing-farm/7d4fb291-8601-4f97-95e6-0ae901812e30
Notice: State is complete and result is: error
Notice: Final state is: failure
Notice: Infra state is: OK
```

which says

```
Tests failed to run. See [logs](https://artifacts.osci.redhat.com/testing-farm/7d4fb291-8601-4f97-95e6-0ae901812e30/#log-links) for more details.

This can be an infrastructure problem, check [status page](https://status.testing-farm.io/) for known outages.
```

which usually means we should just run the test again via the `[citest_bad]` mechanism, or just use the github action page re-run this job if you don't want to re-run all of the failed jobs.

martinpitt:
[citest_bad]

martinpitt:
[citest_bad]

martinpitt:
Superseded by #351

PR #354, opened 2025-05-16
chore: Clean up __mssql_tuned_supported declaration

martinpitt:
Stop using `| bool` as a way to squeeze a value out of an undefined variable. Instead, define a default in vars/main.yml.

There's no need to typecast it, it's (supposed to) always be a bool, and it's internal.

----

Spotted in https://github.com/linux-system-roles/mssql/pull/349/files#r2091064755 ff.

## Summary by Sourcery

Introduce a default declaration for __mssql_tuned_supported and clean up its usage in task conditionals by removing redundant type casting and duplicate definitions.

Enhancements:
- Simplify task conditionals by removing unnecessary '| bool' filters from __mssql_tuned_supported checks.

Chores:
- Define default __mssql_tuned_supported variable as false in vars/main.yml.
- Remove duplicate __mssql_tuned_supported declaration from vars/Suse.yml.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -442,26 +442,26 @@
   package:
     name: tuned-profiles-mssql
     state: present
-  when: __mssql_tuned_supported | bool
+  when: __mssql_tuned_supported
 
 - name: Ensure that the tuned service is started and enabled
   service:
     name: tuned
     state: started
     enabled: true
-  when: __mssql_tuned_supported | bool
+  when: __mssql_tuned_supported
 
 - name: Get the active Tuned profiles
   command: tuned-adm active
   changed_when: false
   register: __mssql_tuned_active_profiles
-  when: __mssql_tuned_supported | bool
+  when: __mssql_tuned_supported
 
 # adding the mssql profile to the end of the list ensures
 # that it overrides conflicting settings in other profiles
 - name: Add mssql to the list of Tuned profiles
   when:
-    - __mssql_tuned_supported | bool
+    - __mssql_tuned_supported
     - '"mssql" not in __mssql_tuned_active_profiles.stdout'
   block:
     - name: Attempt to add mssql to the list of Tuned profiles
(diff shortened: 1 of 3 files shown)


martinpitt:
[citest]

martinpitt:
[citest_bad]

tasks/main.yml
```diff
@@ -442,26 +442,26 @@
...
   register: __mssql_tuned_active_profiles
-  when: __mssql_tuned_supported | bool
+  when: __mssql_tuned_supported
 
 # adding the mssql profile to the end of the list ensures
 # that it overrides conflicting settings in other profiles
 - name: Add mssql to the list of Tuned profiles
   when:
-    - __mssql_tuned_supported | bool
+    - __mssql_tuned_supported
```
richm, line 464: I would prefer not to change these - I would prefer to leave the ` | bool` intact

vars/main.yml
```diff
@@ -66,6 +66,9 @@ __mssql_ha_private_key_dest: /var/opt/mssql/data/{{ mssql_ha_cert_name }}.pvk
 __mssql_ad_kinit_user: >-
   {{ mssql_ad_kerberos_user }}@{{ ad_integration_realm | upper }}
 
+# OS has tuned with the extra "mssql" profile
+__mssql_tuned_supported: false
```
richm, line 70: This will only matter if the role ever supports anything other than the Red Hat family (RHEL, Fedora, CentOS, Rocky, Alma, etc.) and SUSE e.g. if we ever add support for Ubuntu, we won't need to add a `vars/Ubuntu.yml` in order to define `__mssql_tuned_supported` but we'll probably need to add it anyway for other variables.

martinpitt: Ah, Fedora uses vars/RedHat.yml? That works through `ansible_os_family`? OK.

martinpitt:
Ack, fair enough. This doesn't block anything, nor do I have any emotional attachment here. It just came up as a side issue in #349

vars/main.yml
richm: > Ah, Fedora uses vars/RedHat.yml? That works through `ansible_os_family`? OK.

Yes  https://github.com/ansible/ansible/blob/40c919d7bdcadf5e17704995b5950fd78f84da7e/lib/ansible/module_utils/facts/system/distribution.py#L512

PR #369, opened 2025-08-01
test: verify that tests_ha_single fails on EL7 in the beginning

spetrosi:
Enhancement: verify that tests_ha_single fails on EL7 in the beginning

## Summary by Sourcery

Reorder and consolidate single-node HA role tests to improve test flow and placement of primary replica and unsupported version failure checks.

Enhancements:
- Move the test verifying failure when mssql_ha_replica_type is not set to after the default EL <8 failure check
- Relocate the tasks for setting mssql_ha_replica_type fact and asserting unsupported EL7/EL9 version failures to follow the primary-node requirement test

### DIFF
--- a/templates/configure_ag.j2
+++ b/templates/configure_ag.j2
@@ -73,16 +73,16 @@ BEGIN
 {# In the first loop, add primaries at the end and witness at the beginning #}
 {% for item in ansible_play_hosts_all %}
 {%   if hostvars[item]['mssql_ha_replica_type'] == 'primary' %}
-{{ ag_replicas_primary_last.append(item) -}}
+{%     set _ = ag_replicas_primary_last.append(item) %}
 {%   elif hostvars[item]['mssql_ha_replica_type'] == 'witness' %}
-{{ ag_replicas_primary_last.insert(0, item) -}}
+{%     set _ = ag_replicas_primary_last.insert(0, item) %}
 {%   endif %}
 {% endfor %}
 {# In the second loop, add others at the beginning #}
 {% for item in ansible_play_hosts_all %}
 {%   if hostvars[item]['mssql_ha_replica_type']
      not in ag_replicas_primary_last %}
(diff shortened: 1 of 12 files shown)


spetrosi:
[citest]

spetrosi:
[citest]

vars/Suse.yml
```diff
@@ -3,7 +3,7 @@
 __mssql_base_repository: "https://packages.microsoft.com/sles/{{ ansible_facts['distribution_major_version'] | int }}/"
 __mssql_server_repository: "{{ __mssql_base_repository }}mssql-server-{{ mssql_version | int }}/"
 __mssql_client_repository: "{{ __mssql_base_repository }}prod/"
-__mssql_client_packages: "mssql-tools{{ '' if __sqlcmd_ver == '17' else __sqlcmd_ver }}"
+__mssql_client_packages: "mssql-tools{{ '' if __sqlcmd_ver | int == 17 else __sqlcmd_ver }}"
```
richm, line 6: I don't think any of these need to be converted to `int` - in the other cases above too - these are just converting from string to int then back to string.  For the `==` comparison - the string comparison should work.

spetrosi:
[citest]

templates/configure_ag.j2
```diff
@@ -73,16 +73,16 @@ BEGIN
 {# In the first loop, add primaries at the end and witness at the beginning #}
 {% for item in ansible_play_hosts_all %}
 {%   if hostvars[item]['mssql_ha_replica_type'] == 'primary' %}
-{{ ag_replicas_primary_last.append(item) -}}
+{%     set _ = ag_replicas_primary_last.append(item) %}
```
richm, line 76: ok - note that earlier versions of ansible 2.19 did not allow the use of mutator methods such as `append`, `update`, etc. but I guess they have relaxed that restriction.

PR #404, opened 2026-03-19
test: ensure role gathers the facts it uses by having test clear_facts before include_role

richm:
The role gathers the facts it uses.  For example, if the user uses
`ANSIBLE_GATHERING=explicit`, the role uses the `setup` module with the
facts and subsets it requires.

This change allows us to test this.  Before every role invocation, the test
will use `meta: clear_facts` so that the role starts with no facts.

Create a task file tests/tasks/run_role_with_clear_facts.yml to do the tasks
to clear the facts and run the role.  Note that this means we don't need to
use `gather_facts` for the tests.

Some vars defined using `ansible_facts` have been changed to be defined with
`set_fact` instead.  This is because of the fact that `vars` are lazily
evaluated - the var might be referenced when the facts have been cleared, and
will issue an error like `ansible_facts["distribution"] is undefined`.  This is
typically done for blocks that have a `when` condition that uses `ansible_facts`
and the block has a role invocation using run_role_with_clear_facts.yml
These have been rewritten to define the `when` condition using `set_fact`.  This
is because the `when` condition is evaluated every time a task is invoked in the
block, and if the facts are cleared, this will raise an undefined variable error.

## Summary by Sourcery

Introduce a shared test helper that clears Ansible facts before each role invocation and update tests to use it so the mssql role is exercised with explicit fact gathering.

Bug Fixes:
- Prevent test failures caused by lazy evaluation of ansible_facts when facts are cleared between role runs.

Enhancements:
- Centralize role execution in tests via a reusable run_role_with_clear_facts task file that supports tasks_from, public, and failed_when semantics.
- Precompute fact-based conditions with set_fact in tests to avoid undefined variable errors after facts are cleared.
- Simplify test playbooks by dropping redundant gather_facts directives, relying on the role to gather required facts.

Tests:
- Refactor mssql role test playbooks to include the new run_role_with_clear_facts helper instead of calling the role directly, ensuring each test run starts from a clean fact state.


richm:
[citest]

richm:
@spetrosi are these errors expected?  They don't look like they are related to clearing facts.

tests/tasks/run_role_with_clear_facts.yml
```diff
@@ -0,0 +1,37 @@
+---
+# Task file: clear_facts, run linux-system-roles.mssql.
```
spetrosi, line 3: btw do you create this file centrally from .github? If yes let's add a comment that manual edits to this file will be overwritten.

spetrosi (approved):

spetrosi:
There is a bunch of failures that are caused by issues in mssql packages from Microsoft. I reported them and had a reply that they are looking into them but still no result.

However, CentOS-7 tests should pass. [tests_tls](https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/tf_mssql-404_CentOS-7-latest-2.9_20260319-213732/artifacts/tests_tls_2017-ANSIBLE-2.9-test_playbooks_parallel-FAIL.log) fail there, maybe it requires public in every role invocation.

spetrosi (changes requested):
fix tests_tls, should pass on CentOS 7.

richm:
[citest]

richm:
[citest_bad]

spetrosi (dismissed):
fix tests_tls, should pass on CentOS 7.

richm:
All of the centos7 tests pass, but TF is reporting an error for some reason

richm:
[citest_bad]

richm:
[citest]

richm:
[citest]

richm:
[citest]

spetrosi (approved):

PR #409, opened 2026-04-26
feat: add role fingerprints to syslog

richm:
Feature: Add a fingerprint string to the system log to indicate when the role began
successfully, and when the role finished successfully.  The fingerprint string indicates
the role name, a timestamp, and the platform.

Reason: Users can see when the role was used and if it was used successfully.  This
information from the system log can be collected by log scanners and aggregators
for further analysis.

Result: The role logs fingerprints to the system log.

This also adds a test to check if the fingerprints were written upon a successful
role invocation.

## Summary by Sourcery

Add syslog fingerprinting for the mssql role and cover it with an integration-style test.

New Features:
- Introduce an sr_fingerprint Ansible module to write structured fingerprint messages to syslog.
- Log begin and success fingerprint messages when the mssql role starts and completes successfully.

Tests:
- Add an end-to-end test playbook that runs the mssql role and verifies the expected fingerprint entries appear in the system journal.


richm:
@spetrosi I added a tests_default.yml which should always run on each supported platform.  If there is a more appropriate test to add the fingerprint check to I can do that too

richm:
[citest]

tests/tests_default.yml
spetrosi, line 1: Here please use the existing [tests_default_2019.yml](https://github.com/linux-system-roles/mssql/blob/main/tests/tests_default_2019.yml)

richm: I need a test that runs the role successfully.  It looks like tests_default_2019.yml is expected to fail with the EULA error.

spetrosi: Try tests/tests_include_vars_from_parent.yml

spetrosi: The one you did is similar to tests_accept_eula, which fails on mssql variables not defined.

spetrosi:
[citest]

richm:
ah crud - sorry - didn't mean to merge this

PR #412, opened 2026-04-29
feat: add Ubuntu 22.04 and 24.04 support

mhrznamn068:
- Add vars/Debian.yml, vars/Ubuntu_22.yml, vars/Ubuntu_24.yml for OS-specific variable overrides (repo URLs, package names, TLS paths)
- Add SQL Server 2025 (ver17) to version package mapping
- Add apt-based GPG key, server repo, and tools repo setup in tasks/main.yml
- Fix hardcoded /etc/pki/tls paths to use configurable vars so TLS works on both RHEL (/etc/pki/tls) and Ubuntu (/etc/ssl)
- Fail fast with clear messages when mssql_ha_configure or mssql_ad_configure is used on Ubuntu (unsupported)
- Update meta/main.yml to declare Ubuntu 22.04 and 24.04 as supported

## Summary by Sourcery

Add Ubuntu 22.04 and 24.04 support for the SQL Server Ansible role, including package repositories, TLS paths, and version mappings, while explicitly limiting unsupported HA and AD features on Ubuntu.

New Features:
- Declare Ubuntu 22.04 and 24.04 as supported platforms for the SQL Server role.
- Introduce Debian/Ubuntu-specific variables for repositories, TLS paths, and client package selection.
- Support SQL Server 2025 through an updated version-to-package mapping.

Bug Fixes:
- Make TLS certificate and key destinations configurable to work across both RHEL-based and Debian-based systems.

Enhancements:
- Add apt-based handling for Microsoft GPG keys and repository configuration on Ubuntu, including upgrade handling and cache updates.
- Prevent unsupported mssql_ha_configure and mssql_ad_configure operations on Ubuntu with explicit fail-fast tasks and clear messages.

## Summary by Sourcery

Add Ubuntu 22.04 and 24.04 support to the SQL Server Ansible role, including Debian-family variable overrides, repository setup, TLS path configurability, and updated version mapping.

New Features:
- Support Ubuntu 22.04 and 24.04 as target platforms for the SQL Server role via dedicated Debian/Ubuntu variable files and Galaxy metadata.
- Introduce Debian/Ubuntu-specific repository and client configuration, including per-version SQL Server repositories and tools repositories.
- Add support for SQL Server 2025 through an extended version-to-package mapping.

Bug Fixes:
- Make TLS certificate and key locations configurable so TLS works correctly across RHEL-based and Debian-based systems.

Enhancements:
- Add apt-based handling for Microsoft GPG keys and Microsoft SQL Server/server tools repositories on Ubuntu, including upgrade handling and apt cache updates.
- Ensure TLS private key directories are created with appropriate permissions before configuring encryption.
- Fail fast with clear messages when unsupported HA or AD configuration options are requested on Ubuntu.
- Standardize shell task execution by explicitly using bash where required.


tasks/main.yml
```diff
@@ -345,6 +395,52 @@
...
+- name: Prepare for the upgrade on Ubuntu
+  when:
+    - __mssql_current_version is defined
+    - mssql_upgrade | bool
+    - ansible_facts["pkg_mgr"] == 'apt'
+    - __mssql_server_repository_list_url is defined
+    - __mssql_current_version | int != mssql_version | int
+  file:
+    path: /etc/apt/sources.list.d/mssql-server-{{ __mssql_current_version | int }}.list
+    state: absent
```
spetrosi, line 418: This task is not idempotent and will run each role invocation. So each time it will remove this file and download it again. Is it possible to verify e.g. checksums of this file and the file to be installed?

tasks/main.yml
```diff
@@ -345,6 +395,52 @@
...
+      get_url:
+        url: "{{ __mssql_prod_pkg_url }}"
+        dest: /tmp/packages-microsoft-prod.deb
+        mode: '0644'
+
+    - name: Install packages-microsoft-prod.deb
+      apt:
+        deb: /tmp/packages-microsoft-prod.deb
+
+- name: Configure the Microsoft SQL Server repo on Ubuntu 22 using list file
```
spetrosi: Can this be part of the block above? Conditions looks the same

tasks/main.yml
```diff
@@ -580,6 +692,16 @@
...
   block:
+    - name: Ensure TLS private key directory exists
+      file:
+        path: "{{ __mssql_tls_private_key_dest_dir }}"
+        state: directory
+        owner: mssql
+        group: mssql
+        mode: "0700"
+      when:
+        - mssql_tls_enable | bool
```
spetrosi:
```suggestion
      when: mssql_tls_enable | bool
```

vars/main.yml
```diff
@@ -33,6 +33,9 @@ __mssql_version_package_mapping:
   - 2017: 14
   - 2019: 15
   - 2022: 16
+  - 2025: 17
```
spetrosi, line 36: I didn't know there is a new version. Is it fully supported?

mhrznamn068: Yes 2025 version is supported, just configured using this branch of mine in Ubuntu 24.04

tasks/main.yml
```diff
@@ -249,6 +267,38 @@
...
+        mode: '0644'
+      register: __mssql_gpg_download
+      until: __mssql_gpg_download is success
+      retries: 3
+      delay: 5
+
+    - name: Convert Microsoft GPG key to binary keyring format
+      command: >-
+        gpg --batch --no-tty --yes --dearmor -o {{ __mssql_gpg_key_dest }} /tmp/microsoft.asc
+      changed_when: __mssql_gpg_download is changed
```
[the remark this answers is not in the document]
mhrznamn068: fixed in d7dee7d4346a787a9f49674c1bfc783e32b81dd0 commit

tasks/main.yml
mhrznamn068: Fixed by downloading to a temp file first, then using copy with remote_src: true to compare checksums before writing subsequent runs report ok when the .list content hasn't changed. Temp file is cleaned up after.

tasks/main.yml
mhrznamn068: Good point. The conditions weren't exactly the same (__mssql_prod_pkg_url is defined vs __mssql_server_repository_list_url is defined) but the shared outer conditions are identical. Merged into a single block with the shared conditions at the block level and the URL-specific checks moved down to individual tasks.

tasks/main.yml
```diff
@@ -249,6 +267,49 @@
...
+
+    - name: Check if Microsoft GPG keyring already exists
+      stat:
+        path: "{{ __mssql_gpg_key_dest }}"
+      register: __mssql_gpg_key_stat
+
+    - name: Convert Microsoft GPG key to binary keyring format
+      command: >-
+        gpg --batch --no-tty --yes --dearmor -o {{ __mssql_gpg_key_dest }} /tmp/microsoft.asc
+      when: __mssql_gpg_download is changed or not __mssql_gpg_key_stat.stat.exists
```
spetrosi, line 299: if this stat exists this whole block can be skipped. So put this task above the block and make the block run if this doesn't exist.
Otherwise, if the key already exists, it's still got downloaded and removed.

tasks/main.yml
```diff
@@ -249,6 +267,49 @@
...
+    - name: Ensure gnupg is installed for key conversion
+      apt:
+        name: gnupg
+        state: present
+        update_cache: true
+
+    - name: Download Microsoft GPG key
+      get_url:
+        url: "{{ mssql_rpm_key }}"
+        dest: /tmp/microsoft.asc
```
spetrosi, line 284: Please use `tempfile` for such operation. Like in https://github.com/linux-system-roles/mssql/blob/main/tasks/sqlcmd_input_file.yml#L5

tasks/main.yml
spetrosi: `__mssql_gpg_download is changed` how it can be not changed?

tasks/main.yml
```diff
@@ -345,6 +406,63 @@
...
+  when:
+    - ansible_facts["pkg_mgr"] == 'apt'
+    - >-
+      (__mssql_server_packages not in ansible_facts.packages) or
+      (mssql_upgrade | bool)
+  block:
+    - name: Download packages-microsoft-prod.deb
+      get_url:
+        url: "{{ __mssql_prod_pkg_url }}"
+        dest: /tmp/packages-microsoft-prod.deb
```
spetrosi, line 430: use tempfile too. That's for safety to not overwrite files that might exist in users' environment

tasks/main.yml
```diff
@@ -345,6 +406,63 @@
...
+
+    - name: Install packages-microsoft-prod.deb
+      apt:
+        deb: /tmp/packages-microsoft-prod.deb
+      when: __mssql_prod_pkg_url is defined
+
+    - name: Download SQL Server repository list file to temp location
+      get_url:
+        url: "{{ __mssql_server_repository_list_url }}"
+        dest: /tmp/mssql-server-{{ mssql_version | int }}.list
```
spetrosi, line 442: here tempfile too

tasks/main.yml
```diff
@@ -345,6 +406,63 @@
...
+    - name: Download packages-microsoft-prod.deb
+      get_url:
+        url: "{{ __mssql_prod_pkg_url }}"
+        dest: /tmp/packages-microsoft-prod.deb
+        mode: '0644'
+      when: __mssql_prod_pkg_url is defined
+
+    - name: Install packages-microsoft-prod.deb
+      apt:
+        deb: /tmp/packages-microsoft-prod.deb
```
spetrosi, line 436: Actually, ansible-doc for `apt` shows an example as such, sounds like you don't need to download this file:
```
- name: Install a .deb package from the internet
  ansible.builtin.apt:
    deb: https://example.com/python-ppq_0.1-1_all.deb

```

tasks/main.yml
```diff
@@ -345,6 +406,63 @@
...
+        url: "{{ __mssql_server_repository_list_url }}"
+        dest: /tmp/mssql-server-{{ mssql_version | int }}.list
+        mode: '0644'
+        force: true
+      when: __mssql_server_repository_list_url is defined
+
+    - name: Install SQL Server repository list file only if content changed
+      copy:
+        src: /tmp/mssql-server-{{ mssql_version | int }}.list
+        dest: /etc/apt/sources.list.d/mssql-server-{{ mssql_version | int }}.list
```
spetrosi, line 450: Please make this idempotent by first identifying if this file exists and has a correct content. If not exist the role can download directly to this path. If exists - check if content is correct. If not correct - move the original file to a backup location and download a new file here.

tasks/main.yml
```diff
@@ -570,6 +690,21 @@
   failed_when: not __zypper_addrepo_tools_output.rc in [0, 4]
   changed_when: __zypper_addrepo_tools_output.rc != 4
 
+- name: Configure the Microsoft SQL Server Tools repository on Ubuntu 22 using list file
+  get_url:
+    url: "{{ __mssql_client_repository_list_url }}"
+    dest: /etc/apt/sources.list.d/mssql-release.list
```
spetrosi, line 696: can you make this idempotent too?

spetrosi (commented):
Thanks for applying my suggestions, here is another review

meta/main.yml
```diff
@@ -20,6 +20,10 @@ galaxy_info:
     - name: opensuse
       versions:
         - "15.5"
+    - name: Ubuntu
```
spetrosi, line 23: ansible-lint doesn't work well with these definitions. You can define ubunthu versions in galaxy_tags below

spetrosi:
@mhrznamn068 any update on this?

PR #428, opened 2026-08-12
feat: Support SQL Server 2025 on RHEL 9 and RHEL 10

spetrosi:
Enhancement: Support SQL Server 2025 on RHEL 9 and RHEL 10.

Reason: Microsoft release SQL Server 2025 for RHEL 9 and 10 in February 2026, the role should support this.

Result: You can install SQL Server 2025 on RHEL 9, and newly on RHEL 10.

## Summary by CodeRabbit

* **New Features**
  * Added support for SQL Server 2025, including package signing and upgrade workflows.
  * Added support for RHEL/EL 10 and expanded confined SQL Server support to EL 9 and later.
  * Added SQL Server 2025 coverage for high availability, TLS, firewall, SELinux, password, and idempotency scenarios.

* **Documentation**
  * Updated supported-platform and version guidance, including EL 10 restrictions and SQL Server 2025 signing-key configuration.
  * Documented an RHEL 10 SELinux policy issue and available workarounds.

* **Tests**
  * Added comprehensive SQL Server 2025 installation, upgrade, configuration, and validation coverage.
  * Improved cleanup handling across existing test scenarios.


tasks/main.yml
```diff
@@ -243,10 +236,14 @@
...
   with_dict: "{{ __mssql_version_package_mapping }}"
 
-- name: Deploy the GPG key for Microsoft repositories
+- name: Deploy the GPG keys for Microsoft repositories
   rpm_key:
-    key: "{{ mssql_rpm_key }}"
+    key: "{{ item }}"
     state: present
+  # SQL Server 2025 and later packages are signed with a separate GPG key
+  # Original mssql_rpm_key is required to isntall mssql-tools even on 2025
```
richm:
```suggestion
  # Original mssql_rpm_key is required to install mssql-tools even on 2025
```

tests/playbooks/tests_ad_integration_w_keytab.yml
```diff
@@ -76,3 +76,4 @@
 
     - name: Clean up after the role invocation
       include_tasks: ../tasks/cleanup.yml
+      tags: cleanup
```
richm: We usually use the tag `tests::cleanup` in the other system roles

tests/tasks/assert_fail_on_unsupported_ver.yml
```diff
@@ -1,21 +1,15 @@
...
-      ansible_facts['distribution_major_version'] is version('9', '=')) or
-      (ansible_facts['distribution'] in ['Fedora'])) and
-      mssql_version | int != 2022) }}
+# Populate __mssql_supported_versions for the current platform by running the
+# role's set_vars, so this assertion tracks the role's own support matrix
+# instead of duplicating it.
+- name: Determine the SQL Server versions supported on this platform
+  include_role:
+    name: linux-system-roles.mssql
+    tasks_from: set_vars
```
richm: does this work without the `.yml` suffix?

meta/main.yml
```diff
@@ -5,28 +5,14 @@ galaxy_info:
   company: Red Hat, Inc.
   license: MIT
   min_ansible_version: "2.9"
-  platforms:
```
richm, line 8: This doesn't look right to remove the entire `platforms` section?

spetrosi: Ansible-lint was arguing, I guess Ansible doesn't maintain platforms any more hence there is no need to maintain it from our side. 

```
schema[meta]: roles/mssql/meta/main.yml#L1
$.galaxy_info.platforms[1].versions[0] '10' is not one of ['6.1', '7.1', '7.2', 'all']. See https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#using-role-dependencies
schema[meta]: .ansible/collections/ansible_collections/fedora/linux_system_roles/roles/mssql/meta/main.yml#L1
$.galaxy_info.platforms[1].versions[0] '10' is not one of ['6.1', '7.1', '7.2', 'all']. See https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#using-role-dependencies

```

richm: ansible-lint does not like having version `"10"` under 

```yaml
  platforms:
    - name: EL
      versions:
```

which is why we moved to using the tag `el10` under `galaxy_tags`

richm: so the only change to `meta/main.yml` should be to add `el10` under `galaxy_tags`

spetrosi: Then platform won't list EL 10, and tags will. That's a conflict.

richm: > Then platform won't list EL 10, and tags will. That's a conflict.

This is fine, not a conflict.  This is what every other role does e.g. https://github.com/linux-system-roles/auditd/blob/main/meta/main.yml#L15

```yaml
  platforms:
    - name: Fedora
      versions:
        - all
    - name: EL
      versions:
        - "8"
        - "9"
    - name: Debian
      versions:
        - all
  galaxy_tags:
    - audit
    - auditd
    - security
    - el8
    - el9
    - el10
```

spetrosi:
There is an issue with mssql-server on the Microsoft side: on RHEL 10 with SELinux enforcing, running SQL Server 2025 as a confined application fails during mssql-conf setup with Permission denied. The shipped mssql-server-selinux policy module appears incomplete — it defines no file contexts for the runtime directories under /var/opt/mssql (e.g. data and .system, which fall back to usr_t), and mssql_server_t has no rules to write its own directory types. As a result, confined setup is denied out of the box, and it can't be worked around from the role (a restorecon has no target context to apply).

I've let Microsoft know and am waiting for them to resolve this issue.

spetrosi:
[citest]

